PR #15, opened 2014-02-13
Fix the object id conversion to skip 12-byte strings

raymondfeng:
/to @ritch 
/cc @Nikhilesh912


jjhesk:
please release it.. im on ` "loopback-connector-mongodb": "3.0.0-alpha.1",`

jjhesk:
the new issue that uuid does recognized as an object in the query process.

PR #66, opened 2014-11-12
allows for different kind of indexes

amenadiel:
Let's say you had a mongodb model with properties

```
   "properties" : {
 				"id": {
 					"type": "String",
 					"required": false,
 					"id": 1
 				},
 				"geometry": {
 					"type": "Object",
 					"required": false,
 					"index": {
 						"unique": true
 					}
 				}
 			};
```

when you run autoupdate on this model, it will execute 

```
db[collection].ensureIndex({"geometry":1},{"unique":true});
```
the **1** (which stands for a regular index in natural order) is hardcoded. This pull request let's you create other kind of indexes

For example, to create a ["2dsphere" index](http://docs.mongodb.org/manual/tutorial/build-a-2dsphere-index/), the model properties should be 

```
   "properties" : {
 				"id": {
 					"type": "String",
 					"required": false,
 					"id": 1
 				},
 				"geometry": {
 					"type": "Object",
 					"required": false,
 					"index": {
 						"kind": "2dsphere"
 					}
 				}
 			};
```

and this will execute

```
db[collection].ensureIndex({"geometry":"2dsphere"});
```

I used the "kind" attribute to be consistent with loopback-mysql-connector. Besides 2dsphere indexes, you can also create [2d](http://docs.mongodb.org/manual/core/2d/), [text](http://docs.mongodb.org/manual/core/index-text/) and [geoHaystack](http://docs.mongodb.org/manual/core/geohaystack/) indexes.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -717,14 +717,28 @@ MongoDB.prototype.autoupdate = function (models, cb) {
           index = {};
           index[p] = 1; // Add the index key
           if (typeof properties[p].index === 'object') {
-            // The index value should be the options
-            options = properties[p].index;
-            if(options.background === undefined) {
+            // If there is a mongodb key for the index, use it
+            if (typeof properties[p].index.mongodb === 'object') {
+              options = properties[p].index.mongodb;
+              index[p] = options.kind || 1;
+
+              // Backwards compatibility for former type of indexes
+              if (properties[p].index.unique === true) {
+                options.unique = true;
+              }
+
+            } else {
+              // If there isn't an  properties[p].index.mongodb object, we read the properties from  properties[p].index
+              options = properties[p].index;
+            }
+
+            if (options.background === undefined) {
               options.background = true;
             }
+          // If properties[p].index isn't an object we hardcode the background option and check for properties[p].unique
           } else {
             options = {background: true};
-            if(p.unique) {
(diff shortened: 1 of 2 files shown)


bajtos:
@slnode test please

amenadiel:
It fails on Amazon AMI, and it does because in the index comparison the result does not include the expected text index.

The current format for text indexes is present since release 2.6.0, dating Apr 8th 2014. I guess the default yum package available for the Amazon AMI in Jenkins is MongoDB 2.4.x

I will change the text to check for 2dsphere indexes and just skip the text index creation. It doesn't make me specially happy but it should work.

bajtos:
ping @raymondfeng

websquared:
Is fulltext working on the latest version? Or am I missing something?

MongoDB text index successfully created and I can run this directly in mongo, which works fine:
```
db.Job.find({ $text: { $search: "manager" } })
```

However running this API returning ALL results instead:
```
http://mysite.com/api/Jobs?filter={"$text":{"$search":"manager"}}
```

bajtos:
@websquared FWIW, this patch was not released to npmjs yet AFAICT.

acao:
For the curious, this seems to be in the latest release:

https://github.com/strongloop/loopback-connector-mongodb/blob/v1.11.3/lib/mongodb.js#L871

PR #68, opened 2014-11-20
Support for geoNear queries

timosaikkonen:
I wrote basic support for 2d geo queries. It converts GeoPoints to the GeoJSON format that MongoDb uses and applies 2dsphere indexes to geopoint fields when you set index: true on your model definition. See the tests for more details. 

There's a hitch with backwards compatibility though, since thus far geopoints have been saved as {lat: Number, lng: Number} objects, and applying a 2dsphere index on a field like that will fail. Not entirely sure how to deal with that. I suppose one could do a data migration in the autoupdate method?

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -93,6 +93,11 @@ function MongoDB(settings, dataSource) {
     this.settings.enableOptimizedFindOrCreate === true) {
     MongoDB.prototype.findOrCreate = optimizedFindOrCreate;
   }
+  if (this.settings.enableGeoIndexing === true) {
+    MongoDB.prototype.buildNearFilter = buildNearFilter;
+  } else {
+    MongoDB.prototype.buildNearFilter = undefined;
+  }
 }
 
 util.inherits(MongoDB, Connector);
@@ -191,8 +196,41 @@ MongoDB.prototype.fromDatabase = function(model, data) {
         // Convert the Binary into String
         data[p] = data[p].toString();
       }
+    } else if (data[p] && prop && prop.type && prop.type.name === 'GeoPoint' &&
+        this.settings.enableGeoIndexing === true) {
+      data[p] = {
+        lat: data[p].coordinates[1],
+        lng: data[p].coordinates[0],
+      };
     }
   }
+
+  return data;
+};
+
+/*!
+ * Convert JSON to database-appropriate format
+ *
+ * @param {String} model The model name
+ * @param {Object} data The JSON data to convert
+ */
+MongoDB.prototype.toDatabase = function(model, data) {
+  if (this.settings.enableGeoIndexing !== true) {
+    return data;
+  }
+
+  var props = this._models[model].properties;
+
+  for (var p in props) {
+    var prop = props[p];
+    if (data[p] && prop && prop.type && prop.type.name === 'GeoPoint') {
+      data[p] = {
+        coordinates: [data[p].lng, data[p].lat],
+        type: 'Point',
+      };
+    }
+  }
+
   return data;
 };
 
@@ -275,6 +313,9 @@ MongoDB.prototype.create = function(model, data, options, callback) {
     data._id = oid; // Set it to _id
     idName !== '_id' && delete data[idName];
   }
+
+  data = self.toDatabase(model, data);
+
   this.execute(model, 'insert', data, { safe: true }, function(err, result) {
     if (self.debug) {
       debug('create.callback', model, err, result);
@@ -312,6 +353,8 @@ MongoDB.prototype.save = function(model, data, options, callback) {
   var oid = self.coerceId(model, idValue);
   data._id = oid;
   idName !== '_id' && delete data[idName];
+
+  data = self.toDatabase(model, data);
 
   this.execute(model, 'save', data, { w: 1 }, function(err, result) {
     if (!err) {
@@ -459,6 +502,8 @@ MongoDB.prototype.updateOrCreate = function updateOrCreate(model, data, options,
   var idName = self.idName(model);
   var oid = self.coerceId(model, id);
   delete data[idName];
+
+  data = self.toDatabase(model, data);
 
   // Check for other operators and sanitize the data obj
   data = self.parseUpdateData(model, data, options);
@@ -676,6 +721,39 @@ MongoDB.prototype.buildSort = function(model, order) {
   return sort;
 };
 
+function buildNearFilter(query, near) {
+  var coordinates = {};
+  if (typeof near.near === 'string') {
+    var s = near.near.split(',');
+
+    coordinates.lng = Number(s[0]);
+    coordinates.lat = Number(s[1]);
+  } else {
+    coordinates = near.near;
+  }
+
+  query.where[near.key] = {
+    near: {
+      $geometry: {
+        coordinates: [coordinates.lng, coordinates.lat],
+        type: 'Point',
+      },
+    },
+  };
+};
+
+function hasNearFilter(where) {
+  if (!where) return false;
+
+  for (var k in where) {
+    if (where[k] && typeof where[k] === 'object' && where[k].near) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 /**
  * Find matching model instances by the filter
  *
@@ -715,35 +793,39 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
       return callback(err);
     }
     var order = {};
-    if (!filter.order) {
-      var idNames = self.idNames(model);
-      if (idNames && idNames.length) {
-        filter.order = idNames;
-      }
-    }
-    if (filter.order) {
-      var keys = filter.order;
-      if (typeof keys === 'string') {
-        keys = keys.split(',');
-      }
-      for (var index = 0, len = keys.length; index < len; index++) {
-        var m = keys[index].match(/\s+(A|DE)SC$/);
-        var key = keys[index];
-        key = key.replace(/\s+(A|DE)SC$/, '').trim();
-        if (key === idName) {
-          key = '_id';
-        }
-        if (m && m[1] === 'DE') {
-          order[key] = -1;
-        } else {
-          order[key] = 1;
+
+    // don't apply sorting if dealing with a geo query
+    if (!hasNearFilter(filter.where)) {
+      if (!filter.order) {
+        var idNames = self.idNames(model);
+        if (idNames && idNames.length) {
+          filter.order = idNames;
         }
       }
-    } else {
-      // order by _id by default
-      order = { _id: 1 };
+      if (filter.order) {
+        var keys = filter.order;
+        if (typeof keys === 'string') {
+          keys = keys.split(',');
+        }
+        for (var index = 0, len = keys.length; index < len; index++) {
+          var m = keys[index].match(/\s+(A|DE)SC$/);
+          var key = keys[index];
+          key = key.replace(/\s+(A|DE)SC$/, '').trim();
+          if (key === idName) {
+            key = '_id';
+          }
+          if (m && m[1] === 'DE') {
+            order[key] = -1;
+          } else {
+            order[key] = 1;
+          }
+        }
+      } else {
+        // order by _id by default
+        order = { _id: 1 };
+      }
+      cursor.sort(order);
     }
-    cursor.sort(order);
 
     if (filter.limit) {
       cursor.limit(filter.limit);
@@ -901,6 +983,9 @@ MongoDB.prototype.updateAttributes = function updateAttrs(model, id, data, optio
   var oid = self.coerceId(model, id);
   var idName = this.idName(model);
 
+
+  data = self.toDatabase(model, data);
+
   this.execute(model, 'findAndModify', { _id: oid }, [
     ['_id', 'asc'],
   ], data, {}, function(err, result) {
@@ -935,6 +1020,8 @@ MongoDB.prototype.update =
 
     where = self.buildWhere(model, where);
     delete data[idName];
+
+    data = self.toDatabase(model, data);
 
     // Check for other operators and sanitize the data obj
     data = self.parseUpdateData(model, data, options);
@@ -1040,7 +1127,14 @@ MongoDB.prototype.autoupdate = function(models, cb) {
             if (options.background === undefined) {
               options.background = true;
             }
+          } else if (properties[p].type && properties[p].type.name === 'GeoPoint') {
+            var indexType = typeof properties[p].index === 'string' ?
+              properties[p].index : '2dsphere';
           // If properties[p].index isn't an object we hardcode the background option and check for properties[p].unique
+            options.name = 'index' + indexType + p;
+
+            index[p] = indexType;
+            indexList.push({ keys: index, options: options });
           } else {
             options = { background: true };
             if (properties[p].unique) {
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -1,6 +1,10 @@
 // This test written in mocha+should.js
 var semver = require('semver');
 var should = require('./init.js');
+var testUtils = require('../lib/test-utils');
+var async = require('async');
+
+var GeoPoint = require('loopback-datasource-juggler').GeoPoint;
 
 var Superhero, User, Post, PostWithStringId, db;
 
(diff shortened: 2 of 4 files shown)


bajtos:
@ritch ping

lib/mongodb.js
```diff
@@ -449,6 +479,33 @@ MongoDB.prototype.buildWhere = function (model, where) {
...
+    near: {
+      $geometry: {
+        coordinates: [near.near.lng, near.near.lat],
+        type: 'Point'
+      }
+    }
+  };
+
+  if (near.maxDistance) {
+    query.where[near.key].near.$maxDistance = near.maxDistance * 1609.34; // loopback deals with miles by default
```
ritch: I'd like to remove `maxDistance` support until we add it to juggler.

ritch: https://github.com/strongloop/loopback-datasource-juggler/issues/392

ritch:
@timosaikkonen if you can remove the `maxDistance` support we can merge this.

timosaikkonen:
@ritch Done.

ritch:
test please

(this comment is to trigger the ci tests)

voitau:
Any plans to merge this? Thanks.

slnode:

Amir-61:
Not sure why we have not merged this patch.

@ritch , @bajtos Could you please review this PR for final LGTM; it solves the following issues:
https://github.com/strongloop/loopback-connector-mongodb/issues/234
https://github.com/strongloop/loopback-connector-mongodb/issues/236

Thanks :-)

slnode:

Amir-61:
@slnode test please

lib/mongodb.js
```diff
@@ -6,6 +6,7 @@ var util = require('util');
 var async = require('async');
 var Connector = require('loopback-connector').Connector;
 var debug = require('debug')('loopback:connector:mongodb');
+var GeoPoint = require('loopback-datasource-juggler/lib/geo').GeoPoint;
```
bajtos: `loopback-datasource-juggler` may not be require-able from the connector. Can we use a different approach to detect GeoPoint properties? E.g. 

```js
if(prop && prop.type && prop.type.name === 'GeoPoint')
```

lib/mongodb.js
```diff
@@ -170,11 +171,32 @@ MongoDB.prototype.fromDatabase = function (model, data) {
         // Convert the Binary into String
         data[p] = data[p].toString();
       }
+    } else if(prop && prop.type === GeoPoint) {
+      data[p] = new GeoPoint({ lat: data[p].coordinates[1], lng: data[p].coordinates[0] })
```
bajtos: I think it should be ok to return a plain object here, that's what we are already doing in the postgresql connector:
 - https://github.com/strongloop/loopback-connector-postgresql/blob/8837e6d5e2b7210f3a286fb0f715fd7c1936dda8/lib/postgresql.js#L234-L242
 - https://github.com/strongloop/loopback-connector-postgresql/blob/8837e6d5e2b7210f3a286fb0f715fd7c1936dda8/lib/postgresql.js#L465-L470

test/mongodb.test.js
```diff
@@ -1,6 +1,8 @@
 // This test written in mocha+should.js
 var should = require('./init.js');
 
+var GeoPoint = require('loopback-datasource-juggler/lib/geo').GeoPoint;
```
bajtos: This can be simplified to `require('loopback-datasource-juggler').GeoPoint`

bajtos:
@slnode test please

bajtos:
@timosaikkonen oops, looks like we have completely forgotten about this pull request. Are you still keen to work on this? If yes, then could you please rebase your code on the latest master and make sure `npm test` still passes?

mediastuttgart:
@bajtos @timosaikkonen how about adding support for $geoWithin queries? 2dsphere indexes support $geoWithin.

bajtos:
> how about adding support for $geoWithin queries? 2dsphere indexes support $geoWithin. guess this also has to be added to https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/geo.js?
@mediastuttgart please open a new issue to track that feature request. It will allow us to keep this pull request small and focused; and ultimately get it landed sooner.

timosaikkonen:
@bajtos Rebased, ran tests, and hit something of a wall. `loopback-datasource-juggler/test/persistence-hooks.suite.js` does a couple of geo queries without declaring a `GeoPoint` property in `TestModel`. This makes the actual Mongo query fail. I can't really see how to get past this without fixing the test in datasource-juggler first. How would you suggest we solve this?

bajtos:
@timosaikkonen Interesting. Could you please point me to the source code of the failing tests? 

I found this test failure in the CI output:

```
Persistence hooks PersistedModel.find triggers `access` hook for geo queries:
     MongoError: Unable to execute query: error processing query: ns=lb-ds-mongodb-test-1022.TestModel limit=1000 skip=0
Tree: GEONEAR  field=geo maxdist=1.79769e+308 isNearSphere=0
Sort: {}
Proj: {}
 planner returned error: unable to find index for $geoNear query
```

and a more useful message from a different MongoDB version

```
 1) Persistence hooks PersistedModel.find triggers `access` hook for geo queries:
     MongoError: can't find any special indices: 2d (needs index), 2dsphere (needs index),  for: { geo: { $near: { $geometry: { coordinates: [ 10, 20 ], type: "Point" } } } }
```

Do I understand it correctly that with your change in place, it's no longer possible to query geopoint properties that are not indexed? Or is it the failing test case exercising a scenario that does not make sense and would not be encountered in real world?

We are quite careful about introducing breaking changes. Can we "hide" this new geo-query feature behind a flag that will be turned off by default to preserve backwards compatibility? I think that will also solve the concern you mentioned in the PR description:

> There's a hitch with backwards compatibility though, since thus far geopoints have been saved as {lat: Number, lng: Number} objects, and applying a 2dsphere index on a field like that will fail. Not entirely sure how to deal with that. I suppose one could do a data migration in the autoupdate method?

@timosaikkonen @raymondfeng @ritch @Amir-61 thoughts?

ahmed-abdulmoniem:
Any news about this PR guys?

near filter is not working and it is blocking us?

ahmed-abdulmoniem:
???????

timosaikkonen:
@bajtos Okie dokes, I hid the functionality behind the `enableGeoIndexing` config flag. Tests run fine now.

lib/mongodb.js
```diff
@@ -191,8 +196,38 @@ MongoDB.prototype.fromDatabase = function(model, data) {
         // Convert the Binary into String
         data[p] = data[p].toString();
       }
+    } else if (data[p] && prop && prop.type && prop.type.name === 'GeoPoint' &&
+      this.settings.enableGeoIndexing === true) {
```
bajtos: Please indent by two levels, otherwise it's difficult to distinguish what's part of the `if` condition and where the branch body begins.

```js
    } else if (data[p] && prop && prop.type && prop.type.name === 'GeoPoint' &&
        this.settings.enableGeoIndexing === true) {
      data[p] = // ...
    }

File: lib/mongodb.js [hunk 6]
```diff
@@ -191,8 +196,38 @@ MongoDB.prototype.fromDatabase = function(model, data) {
...
+ * @param {String} model The model name
+ * @param {Object} data The JSON data to convert
+ */
+MongoDB.prototype.toDatabase = function(model, data) {
+  var props = this._models[model].properties;
+
+  for (var p in props) {
+    var prop = props[p];
+    if (data[p] && prop && prop.type && prop.type.name === 'GeoPoint' &&
+      this.settings.enableGeoIndexing === true) {
```
bajtos: Ditto.

In fact, it will be better to extract this long condition into a private method:

```js
    if (this._shouldConvertToCoordinates(data[p], prop)) {
      data[p] = // ...
    }
```

And the implementation:

```js
MongoDB.prototype._shouldConvertToCoordinates = function(data, prop) {
  return data[p] && prop && prop.type && prop.type.name === 'GeoPoint' &&
    this.settings.enableGeoIndexing === true;
};
```

lib/mongodb.js
```diff
@@ -1040,7 +1124,15 @@ MongoDB.prototype.autoupdate = function(models, cb) {
             if (options.background === undefined) {
               options.background = true;
             }
+          } else if (properties[p].type && properties[p].type.name === 'GeoPoint') {
+            var indexType =
+              typeof properties[p].index === 'string' ?
+                properties[p].index : '2dsphere';
           // If properties[p].index isn't an object we hardcode the background option and check for properties[p].unique
```
bajtos: The indentation of this line looks off to me, could you please check? It may be a problem of GH diff viewer.

test/mongodb.test.js
```diff
@@ -1309,6 +1313,7 @@ describe('mongodb connector', function() {
     });
   });
 
+
```
bajtos: Extra empty line, please remove.

lib/mongodb.js
```diff
@@ -275,6 +310,9 @@ MongoDB.prototype.create = function(model, data, options, callback) {
     data._id = oid; // Set it to _id
     idName !== '_id' && delete data[idName];
   }
+
+  data = self.toDatabase(model, data);
```
bajtos: I find it a bit suspicious, why it was not necessary to call `toDatabase` before? @superkhau @raymondfeng you are more familiar with this code base, do you know the answer?

bajtos:
@superkhau @raymondfeng @Amir-61 PTAL, you are more familiar with connectors than I am. If you don't find any issues (or do not comment at all), I'll land this patch soon after my comments above get addressed.

lib/mongodb.js
timosaikkonen: Actually there's no reason to iterate through the props at all if `enableGeoIndexing !== true`. So I'll just go ahead and check the flag and return at the top of the method.

lib/mongodb.js
timosaikkonen: Really depends on how you like your ternary conditionals indented, but point taken. I'll change it to a less....controversial indentation.

lib/mongodb.js
superkhau: This `toDatabase` is implemented by this PR (line 217). It looks like the user is massaging the data into a [usable format for mongodb](https://docs.mongodb.org/manual/reference/command/geoNear/#command-syntax). We didn't have it before before because we didn't need to pass in the coordinate pair before so it LGTM. @raymondfeng should have the final say though.

@raymondfeng ^

superkhau:
@bajtos I left a comment, it LGTM.

ahmed-abdulmoniem:
@timosaikkonen After using this patch .. where we should use `enableGeoIndexing`?

timosaikkonen:
@ahmed-abdulmoniem `enableGeoIndexing` goes in your MongoDB datasource config:
```json
"mongodb": {
    "host": "mongo",
    "port": 27017,
    "database": "test",
    "name": "mongodb",
    "connector": "loopback-connector-mongodb",
    "enableGeoIndexing": true
  }
```

You'll then also have to declare your location properties as GeoPoints.

ahmed-abdulmoniem:
@timosaikkonen Thanks. This way `near` filter will be working right?

timosaikkonen:
@ahmed-abdulmoniem Yup.

Amir-61:
LGTM

@raymondfeng,
@bajtos and @superkhau gave their LGTM as well. However based on [this comment](https://github.com/strongloop/loopback-connector-mongodb/pull/68#discussion_r60287031) it seems the final LGTM needs to be from you. Please let us know if it LGTY and we can land this patch.

Thanks!

ahmed-abdulmoniem:
@raymondfeng?

ahmed-abdulmoniem:
@raymondfeng This pull request opened in 2014 .. can you please take a look?

ahmed-abdulmoniem:
@superkhau Can you take the lead on that and push the change? I think @raymondfeng is not here?

bajtos:
Landed, thank you for the contribution! And sorry for the long wait :(

I have published `loopback-connector-mongodb@1.14.0` containing this new feature, enjoy!

leanelias:
In version 1.14 not contains enableGeoIndexing

superkhau:
@bajtos Good call. No point waiting if we don't get responses, just fix it later with another PR if anything. @ahmed-abdulmoniem Thanks for the continual follow up. ;) We have new processes in place now to get community PRs prioritized (so we can get them merged in quicker), so hopefully this long turnaround time doesn't happen too often in the future. ;)

superkhau:
@bajtos mongo connector is listing 1.13.2 FYI (should be 1.14.0), https://github.com/strongloop/loopback-connector-mongodb/blob/master/package.json#L3.

@leanelias Looking into this.

superkhau:
@leanelias I see the settings here: https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L96. What issue are you running into? Can you describe it in more detail?

leanelias:
in release 1.14 [](https://github.com/strongloop/loopback-connector-mongodb/blob/d94f835b224c202122bece657c2f1685cc75ba74/lib/mongodb.js#L96). not contains enableGeoIndexing.

ahmed-abdulmoniem:
@superkhau @bajtos Good call. Thank you for the great effort.

bajtos:
Sorry for the confusion, I have accidentally published `1.14.0` from a feature branch. I have fixed the issue by publishing `1.15.0` from the latest master, please give it a try.

zanemcca:
There is a bug with this option. I connect to multiple mongo databases so if the option is not on all them then it will not work on any of them.
This is because the constructor is adding and removing the `buildNearFilter` prototype [here](https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L110).

ahmed-abdulmoniem:
@timosaikkonen Why we removed `maxDistance` support?

ahmed-abdulmoniem:
؟؟

timosaikkonen:
@ahmed-abdulmoniem `maxDistance` support was removed as per @ritch 's request https://github.com/strongloop/loopback-connector-mongodb/pull/68#issuecomment-70534499

ahmed-abdulmoniem:
So it is not working?!!

Actually `near` is not getting accurate results at all!

ahmed-abdulmoniem:
@ritch Can you tell us why we removed the support of `maxDistance`?

geruiz:
@ahmed-abdulmoniem I have the same question. I need maxDistance works  :-(

ahmed-abdulmoniem:
@ritch

geruiz:
@ritch @ahmed-abdulmoniem  It's work with the last version! I used this code: 
Model.find( { 
                where: { 
                      originLocation: {
                              near:  {
                                    $geometry:  {type: "Point", coordinates: [ lng, lat]},
                                    $maxDistance: 50
                                    }
                                }
                         }
                },  callback);
Run well with near and nearSphere.  But geometry attribute need a GeoJSON point, not a GeoPoint from loopback.

seanbrookes:
Hey guys,

I'm also having some trouble getting this working.
- I have v1.15.2 of the mongo connector
- enableGeoIndexing = true
- I have a model named 'Ask' with a property named 'position' that has a 2dShere index on it
- the model json property type is 'GeoPoint'

I run the following find filter in Explorer:  
```
          {"where": {
            "position": {
              "near": {
                "$geometry": {
                  "type": "Point", 
                  "coordinates": [-119.4961611, 49.8876629]
                 }
              }
            }
         }
      }

```
But I keep getting an error back complaining the coordinates are null:
```
{
  "error": {
    "name": "MongoError",
    "status": 500,
    "message": "$near requires a point, given { coordinates: [ null, null ], type: \"Point\" }",
    "$err": "$near requires a point, given { coordinates: [ null, null ], type: \"Point\" }",
    "code": 16885,
    "stack": "..."
  }
}
```

When I run it through the client with DEBUG enabled:
```
  loopback:connector:mongodb all +15ms ask { where: { position: { near: [Object] } } } { [MongoError: $near requires a point, given { coordinates: [ null, null ], type: "Point" }]
  name: 'MongoError',
  message: '$near requires a point, given { coordinates: [ null, null ], type: "Point" }',
```

ahmed-abdulmoniem:
@seanbrookes The point should be GeoJSON format:

```
"location" : {
            "type" : "Point", 
            "coordinates" : [
                31.3452131, 
                30.0620833
            ]
        }
```

seanbrookes:
thanks @ahmed-abdulmoniem, it wasn't clear that creating a model instance with property type geopoint would transform the data structure to geoJSON.

I'm still having trouble with maxDistance though.  It seems the loopback method is different from the mongodb functionality.

Tuxity:
I'm having the same problem as you @seanbrookes,
Anyone got the `maxDistance` working ?

MB-Hilo:
@timosaikkonen hi, do you have an example of near query that uses "enableGeoIndexing"? When I set "enableGeoIndexing": true, I'm getting an error during the query that says 

>  "stack": "TypeError: Cannot read property 'lng' of undefined\n    at MongoDB.buildNearFilter (C:\\Users\\Hilo\\Documents\\parkme\\parkit API\\parkit-API\\node_modules\\loopback-connector-mongodb\\lib\\mongodb.js:791:34)\n    at Function.find (C:\\Users\\Hilo\\Documents\\parkme\\parkit API\\parkit-API\\node_modules\\loopback-datasource-juggler\\lib\\dao.js:1942:17)\n    at Function.Ticket.near (C:\\Users\\Hilo\\Documents\\parkme\\parkit API\\parkit-API\\common\\models\\geoparkingticket.js:59:14)\n"

My query works and returns results when the enableGeoIndexing is false. Wondering what my query format should be to get queries with an index.

`	  var here = new loopback.GeoPoint({lat:lat, lng:lng});

      Ticket.find( {"where":{
      		location:{
	      		near: here, 
	      		maxDistance: 0.1, 
	      		unit: 'kilometers'
	      	}
	      	}}, callback); 

	};`

timosaikkonen:
Hey @MB-Hilo, sorry, I don't. I haven't touched loopback at all in a couple of years.

PR #70, opened 2014-11-30
Make all mongodb update operators available

alexandrusavin:
Only $set operator was available through the update method of the mongodb connector. This adds the ability to use also '$inc', '$max', '$min', '$mul', '$rename', '$setOnInsert', '$set', '$unset'. It falls back to $set if no operator is used which should not break the current behaviour.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -645,7 +645,7 @@ MongoDB.prototype.update =
     var parsedData = {};
     delete data[idName];
 
-    if (self.settings.extended_operators === true) {
+    if (self.settings.allowExtendedOperators === true) {
       var acceptedOperators = [
         // Field operators
         '$currentDate', '$inc', '$max', '$min', '$mul', '$rename', '$setOnInsert', '$set', '$unset',
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -637,18 +637,27 @@ MongoDB.prototype.update =
       debug('updateAll', model, where, data);
     }
     var idName = this.idName(model);
-    delete data[idName];
 
     where = self.buildWhere(model, where);
 
     var acceptedOperators = ['$currentDate', '$inc', '$max', '$min', '$mul', '$rename', '$setOnInsert', '$set', '$unset'];
-    var usedOperator = Object.keys(data)[0];
 
-    if (acceptedOperators.indexOf(usedOperator) === -1) {
-      data.$set = data;
+    var usedOperators = Object.keys(data).filter(function (item) {
+      return acceptedOperators.indexOf(item) !== -1;
+    });
+
+    var parsedData={};
+    if (usedOperators.length === 0) {
+      delete data[idName];
+      parsedData.$set = data;
+    } else {
+      delete data[usedOperators[0]][idName];
+      // Clean the data object and leave only the first found operator
+      //(see https://github.com/strongloop/loopback-connector-mongodb/pull/70#issuecomment-68143689)
+      parsedData[usedOperators[0]] = data[usedOperators[0]];
     }
 
-    this.collection(model).update(where, data, {multi: true, upsert: false},
+    this.collection(model).update(where, parsedData, {multi: true, upsert: false},
       function (err, count) {
         if (self.debug) {
           debug('updateAll.callback', model, where, data, err, count);
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -640,7 +640,16 @@ MongoDB.prototype.update =
     delete data[idName];
 
     where = self.buildWhere(model, where);
-    this.collection(model).update(where, {$set: data}, {multi: true, upsert: false},
+
+    var acceptedOperators = ['currentDate', 'inc', 'max', 'min', 'mul', 'rename', 'setOnInsert', 'set', 'unset'];
+    var testOperator = new RegExp("^\\$("+acceptedOperators.join("|")+")$");
+    var usedOperator = Object.keys(data)[0];
+
+    if (!testOperator.test(usedOperator)) {
+      data.$set = data;
+    }
+
+    this.collection(model).update(where, data, {multi: true, upsert: false},
       function (err, count) {
         if (self.debug) {
           debug('updateAll.callback', model, where, data, err, count);
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -493,7 +493,7 @@ MongoDB.prototype.all = function all(model, filter, callback) {
     if (typeof keys === 'string') {
       keys = keys.split(',');
     }
-    for (var index in keys) {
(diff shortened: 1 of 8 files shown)


raymondfeng:
@alexandrusavin @amenadiel Let's consider #70 & #72 together. Would you please do the following?

- Rebase the patches off the latest master. 
- Introduce a configuration property for the mongodb connector to control if extended operators are allowed.

alexandrusavin:
I'm actually working right now on the update from @amenadiel and to add some more tests. I'll commit my current changes, rebase, add the config property and than add the support for the other operators from #72.

How does that sound? I hope to finish it over the weekend.

amenadiel:
@raymondfeng @alexandrusavin let's agree on the configuration property. 

Should it be set ad dataSource level, for instance to have granular control over which datasources allow what?

```
{
    "user_datasets": {
		"url": "mongodb:/database:27017/user_datasets?w=0",
		"name": "user_datasets",
		"connector": "mongodb",
                "extended_operators": true
    },
    "public_datasets": {
		"url": "mongodb:/database:27017/public_datasets?w=0",
		"name": "public_datasets",
		"connector": "mongodb",
                "extended_operators": false
    },
}
```

I don't know where else could it be defined, since the idea would be to control de connector from the app, without altering the dependencies in node_modules.

raymondfeng:
Yes, it should be at the datasource level. Each datasource is backed by an instance of connector instantiated with the properties from the ds configuration.

alexandrusavin:
@amenadiel @raymondfeng  I have just pushed the change to accept a config option and added also the array update operators. Can anyone check it?

I still have to add some more tests for the array operators.

lib/mongodb.js
```diff
@@ -640,10 +640,39 @@ MongoDB.prototype.update =
...
     }
     var idName = this.idName(model);
-    delete data[idName];
 
     where = self.buildWhere(model, where);
-    this.collection(model).update(where, {$set: data}, {multi: true, upsert: false},
+    var parsedData = {};
+    delete data[idName];
+
+    if (self.settings.extended_operators === true) {
```
raymondfeng: The property name can be `allowExtendedOperators` to be consistent.

alexandrusavin:
Done

lib/mongodb.js
```diff
@@ -640,10 +640,39 @@ MongoDB.prototype.update =
...
+
+    if (self.settings.allowExtendedOperators === true) {
+      var acceptedOperators = [
+        // Field operators
+        '$currentDate', '$inc', '$max', '$min', '$mul', '$rename', '$setOnInsert', '$set', '$unset',
+        // Array operators
+        '$addToSet', '$pop', '$pullAll', '$pull', '$pushAll', '$push',
+        // Bitwise operator
+        '$bit'
+      ];
```
raymondfeng: Maybe even better for `allowExtendedOperators` to be `true` or an array of operator names.

amenadiel:
@alexandrusavin don't worry about those. I already have test for [ '$addToSet', '$pop', '$pullAll', '$pull', '$pushAll', '$push'] for methods updateOrCreate and updateAttributes

alexandrusavin:
Ok! Than please advise on how to go on. I think this would for now. I skipped some operators but only because they require some extra input to the query. I guess the basic operators are there now.

raymondfeng:
Please let me when two of you have everything consolidated and ready to go.

amenadiel:
@alexandrusavin @raymondfeng it's settled then, the property is `allowExtendedOperators`. I'll modify my code, rebase and push in the next hour.

My code includes tests for array operators and methods updateOrCreate and updateAttributes, whereas @alexandrusavin's operates on method updateAll, and non-array operators. 

If you decide there's no need to write tests for array operators on method updateAll, then this PR is complete.

alexandrusavin:
@amenadiel @raymondfeng I just committed a small fix but I also refactored the parsing logic to use a method instead of reusing the same code in all of the update operations (I completely missed the other update operations :) ).

I also pushed amenadiel's modifications in my branch to be able to actually test this change and fix the bug. So now it looks like it passes all the tests.

lib/mongodb.js
```diff
@@ -309,6 +309,48 @@ MongoDB.prototype.find = function find(model, id, callback) {
...
+      parsedData.$set = data;
+    }
+  } else {
+    parsedData.$set = data;
+  }
+
+  return parsedData;
+};
+
+/**
```
raymondfeng: Please note the Array filter and map involves closures which have adverse performance impacts. Since this function will be invoked per request, can we optimize them using for-loop?

raymondfeng:
@alexandrusavin @amenadiel Thank you for the great work. 

I added one more comment. Please fix it and we should be ready to go.

lib/mongodb.js
amenadiel: I suggest the following

```
    var userOperators=0;
    // each accepted operator will take its place on parsedData if defined
    acceptedOperators.forEach(function(operator) {
      if(data[operator]) { 
          parsedData[operator] = data[operator];
          userOperators++
      }
    });

    // if parsedData is still empty, then we fallback to $set operator
    if(userOperators === 0) {
      parsedData.$set = data;
    }
```

amenadiel:
@raymondfeng @alexandrusavin my suggestion is to do

```js
  if (self.settings.allowExtendedOperators === true) {
    var acceptedOperators = [
      // Field operators
      '$currentDate', '$inc', '$max', '$min', '$mul', '$rename', '$setOnInsert', '$set', '$unset',
      // Array operators
      '$addToSet', '$pop', '$pullAll', '$pull', '$pushAll', '$push',
      // Bitwise operator
      '$bit'
    ];

    var usedOperators = 0;
    // each accepted operator will take its place on parsedData if defined
    acceptedOperators.forEach(function(operator) {
      if(data[operator]) {
          parsedData[operator] = data[operator];
          usedOperators++;
      }
    });

    // if parsedData is still empty, then we fallback to $set operator
    if(usedOperators === 0) {
      parsedData.$set = data;
    }
  } else {
    parsedData.$set = data;
  }
```

alexandrusavin:
@amendaniel Yes, than map is useless. I was trying something and I forgot it there. I'll update it as soon as I get home.

alexandrusavin:
@amenadiel @raymondfeng I added the change.

amenadiel:
@raymondfeng since @alexandrusavin included my squashed commits into this PR I might as well close mine so we all focus in this one. Do you agree?

raymondfeng:
@amenadiel Please close yours to avoid confusions. Thanks!

raymondfeng:
@alexandrusavin @amenadiel I have landed the PR. Thank you for the great work!
FYI: I had to upgrade our mongodb version to 2.6.x to support $max and $mul (not supported in 2.4.x). 

@crandmck We need to document this enhancement in the mongodb connector.

PR #72, opened 2014-12-19
Allows optional use of $set, $addToSet, $push, $pull, $pullAll and $pop

amenadiel:
At this point, loopback's mongodb connector implementation of  updateOrCreate has the [$set](http://docs.mongodb.org/manual/reference/operator/update/set/) operator hardcoded. For example, given the Product 

```
{id:12345, name:'oldname','price':100, 'historicprice':[90]}
```
the following operation
```
Model.updateOrCreate({
     id:12345, 
     name:'newname'
}, function(err,cb));
```
would result in an updated model with all the original properties, but whose name is now 'newname'.

```
{id:12345, name:'newname','price':100, 'historicprice':[90]};
```

With this pull request, you can still use the current method (this is, passing updateOrCreate an object with the properties you want to update) but you can also refine the syntax to specify what do you want to update (under the $set key) and at the same time manipulate an array using [MongoDB array update operators](http://docs.mongodb.org/manual/reference/operator/update-array/). Currently, there's support for $addToSet, $push, $pull, $pullAll and $pop. 

**An example with [$addToSet](http://docs.mongodb.org/manual/reference/operator/update/addToSet/)** 

Given the aforementiones model, the following operation

```
Model.updateOrCreate({
     id:12345, 
     $set: {name:'newname'},
     $addToSet: {historicprice:95},
}, function(err,cb));
```

will $set  a new name, and will append a new value to 'historicprice' instead of overwriting the existing one.

```
{id:12345, name:'newname','price':100, 'historicprice':[90,95]};
```

This is part of an ongoing effort to implement more native MongoDB functions to properly replace existing apps that use lower level libraries.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -323,12 +323,38 @@ MongoDB.prototype.updateOrCreate = function updateOrCreate(model, data, callback
 
   var id = self.getIdValue(model, data);
   var idName = self.idName(model);
-
+  var parsedData = {};
   var oid = ObjectID(id);
   delete data[idName];
-  this.collection(model).findAndModify({_id: oid}, [
+
+  if (self.settings.allowExtendedOperators === true) {
+    var acceptedOperators = [
+      // Field operators
+      '$currentDate', '$inc', '$max', '$min', '$mul', '$rename', '$setOnInsert', '$set', '$unset',
+      // Array operators
+      '$addToSet', '$pop', '$pullAll', '$pull', '$pushAll', '$push',
+      // Bitwise operator
+      '$bit'
+    ];
+
+    // each accepted operator will take its place on parsedData if defined
+    acceptedOperators.forEach(function(operator) {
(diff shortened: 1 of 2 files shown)


raymondfeng:
I wonder if there should be a configuration property for the mongodb connector to control if these extended operations are allowed. Otherwise, it could be vulnerable to code injection.

amenadiel:
Added allowExtendedOperators parameter and modified tests where it was due.

Also added support for @alexandrusavin operators.

amenadiel:
@raymondfeng it's failing an unrelated test, because it gets `connect ECONNREFUSED 127.0.0.1:4` instead of just `connect ECONNREFUSED`.

This is beyond my control.

raymondfeng:
@amenadiel Don't worry about that.

amenadiel:
squashed and pushed then. Ready to go.

PR #77, opened 2015-01-13
Upgrade to mongodb node driver 2.x

raymondfeng:
/to @ritch or @bajtos 

See http://mongodb.github.io/node-mongodb-native/2.0/meta/changes-from-1.0/

### DIFF
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -88,7 +88,7 @@ describe('mongodb', function () {
       });
       ds.ping(function(err) {
         (!!err).should.be.true;
-        err.message.should.be.equal('failed to connect to [localhost:4]');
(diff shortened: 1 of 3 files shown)


test/mongodb.test.js
```diff
@@ -88,7 +88,7 @@ describe('mongodb', function () {
       });
       ds.ping(function(err) {
         (!!err).should.be.true;
-        err.message.should.be.equal('failed to connect to [localhost:4]');
+        err.message.should.be.equal('connect ECONNREFUSED');
```
bajtos: Please check that the "Test Connection" button converts this new error message into a useful message in the GUI, so that we don't introduce a subtle regression there.

bajtos:
One comment, the rest LGTM.

PR #86, opened 2015-02-02
Support for regex flags in like/nlike operator

arimus:
Per @andrewburgess:  Basically just added the ability to specify Regex flags so that case insensitive search can be done.

Format would be:

```
Post.find({
  where: {
    title: {
      like: 'someth.*',
      options: 'i'
    }
  }
});
```

Will also allow filters to be used in queries like so:

```
?filter={"where":{"title":{"like":"someth.*","options":"i"}}}
```

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -463,7 +463,9 @@ MongoDB.prototype.buildWhere = function (model, where) {
       return;
     }
     var spec = false;
+    var options = null;
     if (cond && cond.constructor.name === 'Object') {
+      options = cond.options;
       spec = Object.keys(cond)[0];
       cond = cond[spec];
     }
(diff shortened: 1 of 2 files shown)


slnode:

esco:
Hi @arimus thanks for the fix. We just ran into this issue. The change you made works great for the Node API but how would this work with the rest api? 

Ex:
```
?filter=[where][title][like]=something // where do we put options?
```

Thanks!

arimus:
@esco That form for the rest API seems to be case-insensitive already if that's what you are trying to accomplish.  As for other flags, I believe you'll need to use the object form above in the query string, like so:

/api/blah?filter={"where":{"title":{"like":"something","options":"i"}}}

You can use both forms, but this form is more flexible as far as I can tell, even though it's a bit more eye wrenching to see it in the URL.  But it's just an API call, so not too big a deal.

raymondfeng:
ok to test

arimus:
@raymondfeng I can't tell why the merge build failed.  Anything I need to fix on my side?

raymondfeng:
@arimus Some of our build machines have lower versions of mongodb installed and they don't support certain operators such as `$mul`. The PR itself is fine.

projectxmaker:
I tried to do 

npm i loopback-connector-mongodb

but found no changes so far, I missed something ?

esco:
@arimus the form for the rest api isn't case insensitive. I was hoping to use the rest form for easier integration with an autocomplete module. It looks like using options is only possible with the json form at the moment.

Thanks.

kwiky:
@esco This works for me :
?filter=[where][title][like]=something&filter=[where][title][options]=i

esco:
@kwiky The rest form didn't work before this patch was merged. You just answered my question about how to pass options though, thanks!

PR #88, opened 2015-02-03
Update should to 4.x

bajtos:
Update `should` to the latest version `4.6.3`, see also https://github.com/strongloop/loopback-datasource-juggler/pull/421

/to @raymondfeng please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,9 @@
   },
   "devDependencies": {
     "loopback-datasource-juggler": "^2.13.0",
-    "should": "~1.2.2",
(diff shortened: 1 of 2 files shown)


bajtos:
On the second thought I decided to land io.js fix via a standalone pull request.

bajtos:
> BTW, I assume that we need to upgrade should dep for all connectors as they import tests from juggler.

Yes, I think that's correct. It would be nice if it was updated automatically, but I don't know of any easy way.

PR #91, opened 2015-02-05
Disable extended tests if mongodb version is lower than 2.6

raymondfeng:
/to @rmg

### DIFF
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -3,7 +3,7 @@ var should = require('./init.js');
 
 var Superhero, User, Post, PostWithStringId, db;
 
-describe('mongodb', function () {
+describe('mongodb connector', function () {
 
   before(function () {
     db = getDataSource();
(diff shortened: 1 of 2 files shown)


test/mongodb.test.js
```diff
@@ -499,149 +499,162 @@ describe('mongodb', function () {
...
-            should.not.exist(err3);
-
-            User.updateAll({name: 'Simon'}, {name: 'Alex'}, function(err, updatedusers) {
-              should.not.exist(err);
-              updatedusers.should.be.equal(1);
+    describe('extended operators', function() {
+      if (process.env.MONGODB_VERSION) {
+        var semver = require('semver');
+        // semver.lt doesn't accept 2.4 (2.4.0 is required)
+        if (semver.satisfies('2.6.0', '>' + process.env.MONGODB_VERSION)) {
```
rmg: I think you mean:
```js
if (!semver.satisfies(process.env.MONGODB_VERSION, '>=2.6.0')) {
```

rmg:
Rather than putting the test directly in the block.. how about:
```js
var describeMongo26 = describe;
if (process.env.MONGODB_VERSION && !semver.satisfies(process.env.MONGODB_VERSION, '>=2.6.0')) {
  describeMongo26 = describe.skip;
}
describeMongo26('extended operators', function() {
  // ... describe block is run or marked as skipped
});
```

rmg:
Or you could do something similar with `it` and `it.skip`, which might remove the need to wrap everything in a `describe` block and change all that whitespace.

test/mongodb.test.js
raymondfeng: I had to use `semver.satisfies('2.6.0', '>' + process.env.MONGODB_VERSION)` as `!semver.satisfies(process.env.MONGODB_VERSION, '>=2.6.0')` throws if process.env.MONGODB_VERSION is `2.4`

raymondfeng:
@rmg PR updated.

test/mongodb.test.js
rmg: I saw the comment, but it wasn't clear that it was related to reversing the arguments. I understand now.

rmg:
I'm still not a fan of how the range check is done because it makes the code very hard to follow, but I also don't want to block this PR based on differences in coding style, so LGTM.

FWIW, the actual version strings being reported are `v2.6.7' and `v2.4.9`, so they won't cause the invalid version specifier problem.

PR #92, opened 2015-02-06
implement optmized findOrCreate

clarkorz:
1. require strongloop/loopback-datasource-juggler#419 to be merged first
2. cause [one test fail](https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js#L1465) due to id generation, see below

The test fail because the order of pictures are different. The *Example* picture is linked to the author, which calls `findOrCreate()`, but other two don't.

1. By default, *mongodb-driver* will take care of `_id` (generate in client process), but `findAndModify` command is different, when it results an upsert, the `_id` is generated by mongodb server(generate in server process).
2. the connector sorts query by `_id` field, the result might be different among environments.

Any thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 coverage
+.idea
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -938,11 +938,11 @@ MongoDB.prototype.ping = function (cb) {
  * Find a matching model instances by the filter or create a new instance
  *
  * @param {String} model The model name
- * @param {Object} data The model instance data
  * @param {Object} filter The filter
+ * @param {Object} data The model instance data
  * @param {Function} [callback] The callback function
  */
-MongoDB.prototype.findOrCreate = function (model, filter, data, callback) {
+MongoDB.prototype.findOrCreate = function (model, filter, data, options, callback) {
   var self = this;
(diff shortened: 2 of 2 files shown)


clarkorz:
> The *Example* picture is linked to the author, which calls `findOrCreate()`, but other two don't.

I think this is the problem. `HasManyThrough#create()` calls `ThroughModel.create()`, not `ThroughModel.findOrCreate()`.

raymondfeng:
ok to test

raymondfeng:
> I think this is the problem. HasManyThrough#create() calls ThroughModel.create(), not ThroughModel.findOrCreate().

@clarkorz Do you want to create a patch? Otherwise, we cannot merge this PR.

raymondfeng:
BTW, loopback-datasource-juggler@2.17.0 is released.

raymondfeng:
ok to test

clarkorz:
@raymondfeng this PR need strongloop/loopback-datasource-juggler#443 to be merged first. And I found [another issue](https://github.com/strongloop/loopback-datasource-juggler/issues/529).

raymondfeng:
ok to test

raymondfeng:
Duplicate of https://github.com/strongloop/loopback-connector-mongodb/pull/139

Issue #108, opened 2015-03-09
Cannot do simple case insensitive where like


srguiwiz:
This is a problem description.  A second comment will describe one workaround/fix, which might be less than ideal.  A third comment will describe a hopefully more acceptable workaround/fix.

Background fact: JavaScript regular expressions don't support flags, modes, modifiers.  I.e. not `(?i)` for case insensitive. JavaScript regular expressions are used in MongoDB.

A problem if trying to enable end-users to find *Jones, Matthew* by searching for *mat*.

If on the server in Node.js you try to give `find` a `new RegExp("mat","i")` that gets garbled into something useless by `dao.js` function `DataAccessObject._coerce` making it a string that doesn't translate back into the same regular expression for `loopback-connector-mongodb`'s `like`.

Apparently impossible to make it work without a change to source code.

For completeness I will describe a proven fix in `dao.js`, but don't think that is ideal, because I cannot be certain of its impact on other database connectors.  Also that fix is limited to Node.js and doesn't work with RESTful.

Then I will describe supporting `(?i)` within `loopback-connector-mongodb`, which will work on the server in Node.js as well as from RESTful.

srguiwiz:
Here is a proven workaround/fix in `dao.js`.  As I do *not* suggest using it, I do not provide a pull request for this one.

In `DataAccessObject._coerce` name the `for` loop:

```javascript
  whereLoop: for (var p in where) {
```

and for `like` do:

```javascript
            case 'like':
            case 'nlike':
              if (val instanceof RegExp) {
                continue whereLoop;
              }
              if (!(typeof val === 'string')) {
```

After which in Node.js you can pass in e.g. to `find` for a `like` a `new RegExp("mat","i")` and get a case insensitive match.

Doesn't solve the problem for RESTful.  And compatibility with other database connectors isn't known.  Not good, but at least nice to know.

srguiwiz:
Better fix here in `loopback-connector-mongodb`'s `lib/mongodb.js` insert:

```javascript
// make { like: "(?i)whatever" } a case insensitive version of { like: "whatever" }
function newRegExp(cond) {
  if (typeof cond === 'string') {
    if (cond.indexOf('(?i)') !== 0) {
      return new RegExp(cond);
    } else {
      return new RegExp(cond.substring(4), 'i');
    }
  } else {
    return new RegExp(cond);
  }
}
```

and change to:

```javascript
      } else if (spec === 'like') {
[...]

PR #109, opened 2015-03-10
Let like: be case insensitive with (?i)

srguiwiz:
Make { like: "(?i)whatever" } a case insensitive version of { like: "whatever" }.

Fix for https://github.com/strongloop/loopback-connector-mongodb/issues/108

### DIFF
--- a/example/model.js
+++ b/example/model.js
@@ -37,6 +37,10 @@ Customer.destroyAll(function (err) {
       Customer.find({where: {'emails.0.label': 'work'}}, function(err, customers) {
         console.log('Customers matched by emails.0.label', customers);
       });
+
+      Customer.find({where: {'name': {like: '(?i)john'}}, function(err, customers) {
+        console.log('Customers matched case insensitively by name', customers);
+      });
       /*
        customer1.updateAttributes({name: 'John'}, function(err, result) {
        console.log(err, result);
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -439,6 +439,19 @@ function idIncluded(fields, idName) {
   return true;
 }
 
+// make { like: "(?i)whatever" } a case insensitive version of { like: "whatever" }
+function newRegExp(cond) {
+  if (typeof cond === 'string') {
+    if (cond.indexOf('(?i)') !== 0) {
+      return new RegExp(cond);
+    } else {
+      return new RegExp(cond.substring(4), 'i');
+    }
+  } else {
+    return new RegExp(cond);
+  }
+}
+
 MongoDB.prototype.buildWhere = function (model, where) {
   var self = this;
   var query = {};
@@ -478,9 +491,9 @@ MongoDB.prototype.buildWhere = function (model, where) {
           return ObjectID(x);
         })};
       } else if (spec === 'like') {
-        query[k] = {$regex: new RegExp(cond, options)};
+        query[k] = {$regex: newRegExp(cond, options)};
       } else if (spec === 'nlike') {
-        query[k] = {$not: new RegExp(cond, options)};
+        query[k] = {$not: newRegExp(cond, options)};
       } else if (spec === 'neq') {
         query[k] = {$ne: cond};
       }


slnode:

bajtos:
@srguiwiz thank you for the pull request. I believe we already support case-insensitive LIKE queries, see https://github.com/strongloop/loopback/issues/877#issuecomment-75957116

> ```js
var pattern = new RegExp('.*'+query+'.*', "i"); /* case-insensitive RegExp search */ 
Post.find({ where: {title: { like: pattern} } }, 
```
> Via the REST API - see strongloop/loopback-connector-mongodb#86
> ```js
?filter={"where":{"title":{"like":"someth.*","options":"i"}}}
```

Is there any particular reason why you cannot use this already supported format?

srguiwiz:
You are right, I had investigated against a version prior to 1.8.0 and/or before options were documented.

I have now verified `{ like: "something", options: 'i' }` also works in the Node API.

This pull request therefore is not necessary.

slgoldberg:
@crandmck @raymondfeng @bajtos @altsang 

Do you guys know how many person-hours it would have saved if this vital information weren't buried in a random thread deep in the bowels of a pull request discussion, but on the [official documentation](https://docs.strongloop.com/display/public/LB/Where+filter) instead?

This is so simple, yet not described anywhere. I have spent many, many hours on this. It's extremely maddening that this simple and clear information wasn't anywhere to be found. I tried every other variant, read all the work-around threads on the Google Group and other places, but .. nothing worked for me, until this very simple solution.

The only reason I never tried this simple solution, was because you explain in gory (but incorrect -- sadly completely incorrect) detail how to do case-insensitive (and partial) search using `regexp`. You spend multiple paragraphs explaining the ins and outs of `regexp`, but none of it is right (I mean, really, none of it is right!).  However, for `like` (the solution that is actually supported and works flawlessly!) you only give 2 sentences (!!) and you make it sound like it's only for mysql:

> The like and nlike (not like) operators enable you to match SQL regular expressions. The regular expression format depends on the backend data source.

There is really woefully little there to explain anything at all about how to use `like` -- nothing at all, just a couple of really useless examples.

The **most common** thing people want to do is case-insensitive, partial text search. Why don't you just show this simple solution in place of those useless examples? And delete all that crap about `regexp` because I mean, it's utterly and completely 100% total crap.  None of the examples for `regexp` work at all, nor do they address (either) the basic most obvious use-case (case-insensitive partial text matching).  You just kind of throw out there, as an afterthought, to add `/i` for insensitive, but you don't mention that it completely 100% does not work with the REST API, but the text there is so confusing that it's impossible to be able to confirm whether you are claiming it should or not (but I assure you, it doesn't).

At least for `mongodb` (which isn't even mentioned in that documentation -- like, literally nowhere on that page does the word even appear), the `like` solution above works absolutely flawlessly and is trivial. Why not document that?  I understand that the `like` expression syntax differs from datasource to datasource, but at least it works! The link to the Mozilla RegExp stuff doesn't help me at all for the stuff I want to do (simple partial search with case insensitivity).

"Why doesn't it work", you ask? Well, because (a) at least when I use them with the stubs generated by `lb-ng` through (I haven't tried isomorphic with this particular project, but I guess it would probably work -- but most people aren't ready to tackle that quite yet and `lb-ng` is awesome for most things), the RegExps that I create (which I create correctly) devolve into `{}`, so I get REST queries like this:

```javascript
{where: {name: {regexp: {}}}
...
{ where: { name: { regexp: {} } } }     // <-- the resulting filter in the actual REST query

```

Great. that does me a whole lot of good.  (Before you tell me, "you left off `filter: {..}`, no I didn't; I'm just omitting it here for brevity (plus, the filter is a parameter to the REST API, not a part of the query, so when I tell you what I saw in the REST API I mean the value of the `filter` parameter.

And (b), the even if somehow I can convince the API to transmit the correctly formatted regexp as a string.  For example, my autocomplete sends this when I type 'foo':

```javascript
var regExp = new RegExp("/.*" + foo + ".*/", "i");
...
{ filter: {where: { name: { regexp: regExp} } } }    // <-- the resulting filter in call to myModel.find(...)
```

Guess what I get back?  Guess?  ...  OK, you probably guessed it:  Bupkis. Zilch. Zippo. Nada:

```javascript
[]
```

No, that's not an error in my markdown.  It's a whole lot of **nothing**.

And you have to believe, I didn't just try that and walk away. I did the dance -- the `LoopBack Hair Pulling Dance` -- read code, read documentation, followed long threads of many people asking the same thing and getting somewhat terse and seemingly irritated responses basically asking everyone to do what I'm doing (pulling my hair out).

I tried to understand how this could be so complicated. It just didn't make sense. I missed this post, of course, until now. But I mean, I spent a lot of time on this. A lot. :-(

Then I figured, I should just try a bunch of things to see if I could just get lucky.  Examples:

```javascript
{ where: { name: { regexp: ".*foo.*", options: "i"} } }
```

and:

```javascript
{ where: { name: { regexp: "/.*foo.*/i"} } }

{ where: { name: { regexp: "foo", options: "g,i"} } }

{ where: { name: { regexp: "foo", options: ["i", "g"] } } }

{ where: { name: { regexp: ".*foo.*"} } }      // (this one really just to have a laugh)

{ where: { name: { regexp: "%foo%/i"} } }      // (this one really just to have a laugh)
```

Even went crazy a little:

```javascript
{ where: { name: { regex: "/.*foo.*/", '$options': "i" } } }
{ where: { name: { regex: "foo" } } }
{ where: { name: { regex: "foo", options: "/g/i" } } }
{ where: { name: { regex: "foo/i", options: "g" } } }
{ where: { name: { regex: "foo", options: "g,i" } } }

```

Got back nothing each time, or if I got back something, it was not what I wanted. :-(  I had to start questioning my own value to society at that point. :-(

Argh. **So many hours wasted** with so many different ideas on this. So much reading of Mozilla and other RegExp documentation. So much wondering "why, oh why is this so frickin' complicated"? And yes, I did it all. I tried every possible option, but of course the common thread I refused to give up was to try to do this through the `lb-ng` generated LoopBack services from AngularJS, through the REST API, not by cheating and going directly to the `Node.js` version of the API.

I did try in quotes, without quotes, with various ways of constructing the RegExp, without using RegExp and just using plain strings. With and without the leading and trailing `/` and on and on. Because in one place the docs make it sound like maybe you just add `/i` to the end of your string, without adding an initial `/`, so of course I had to try that (though I knew better, I mean, I've been using regular expressions longer than many. (I even know what "grep" stands for.)  But I thought, "Is it possible they just randomly decided to do something completely different for the REST API? I mean, it's not such a crazy thought -- the fact that you can't do simple case-insensitive queries and have to jump through hoops and bend over backwards to do this obvious and simple thing through the REST API when it's easy to do through the Node API makes me think maybe they did do something horrible and hacky to get it working. Let's try this..... Oh crap, no, that was wrong. What was I thinking?"  And on and on.

I seriously entertained the idea of just writing a simple PHP-based REST endpoint so I could query it instead, using $http from my angular client instead of my nicely awesome super-sweet auto-generated LoopBack client services.

You know, I mean, seriously, think about it. Someone writing their first AngularJS app using LoopBack to share models between the client and server, are almost invariably going to end up wanting to do the exact same thing as me:  typeahead/autocomplete a form field in the UI based on data in their models exposed over a RESTful service wrapped by `lb-ng`.

So I mean, why not show **that** example or, even if you don't care about AngularJS (which is fine, I've spent 90% of my time on the LoopBack/Node side in my current project, the AngularJS stuff is just the last mile for me), at least share how wonderful this platform is with your users by actually documenting the most common use-cases.

OK, here is all I needed to do, and it would have saved hours. Put this in the docs, please. The stuff from 2014 is just not going to cut it anymore:

Let's say you are using `AngularJS` with `ui-bootstrap`'s `uib-typeahead` component:

```html
<input ng-model="vm.chosenMember"
       uib-typeahead="item as item.name for item in vm.getMembers($viewValue)">
```

```javascript
angular.module(...).controller(.., function(...) {
...
  vm.getMembers = function(val) {
    ...
    var filter = { where: { name: { like: val, options: 'i'} } };
    return Member.find({filter: filter, ... }).$promise;
  };
...
```

Was that so hard? (Note: this isn't a complete solution obviously, because you still have to handle mapping the results back for `uib-typehead` to show the title correctly, etc., but it's 80% and moreover, it shows exactly what you need to do to get case-insensitive partial text matching -- use `like` with `options: 'i'`. Done. Geez. If this had been in the documentation... I would have actually gotten stuff done tonight and I might not be up all night pulling my hair out. :-( :-(

I mean, is this type of clear/simple solution to a common question hard to put in the docs?  Are the examples in the documentation really even in the right ballpark? Who wants to search for `^T`? Really? C'mon, please, please give useful examples in the docs for the most common usage patterns, not just trivial partial examples that leave the rest as an exercise to the reader to read through multiple bugs, group discussions, complex codebases (e.g. noting how in this thread the documentation was finally found in another pull request).  Why isn't this in the docs?  I mean, isn't this one of the absolute topmost questions you get? I found so many hits when I searched for this topic, which of course just made it that much harder to piece it all together. Ugh.

And to the awesome folks who added this feature, thank you, thank you, thank you. Because RegExp does not work at all. :-( Did I mention? It's a huge black hole, stay away when all you want do is one simple query for your case-insensitive autocomplete. :-(

Thanks for listening.

  Steve (exasperated after many hours of `LoopBack hair-pulling` to do one frigging thing, writing this at 3:45am, so I humbly request you take all this with a grain of salt)

crandmck:
Thanks @slgoldberg for your very detailed comment.  I can understand your frustration, and I'm surprised that the regexp docs are incorrect since @superkhau among others in engineering reviewed them.   I'm also surprise the issue never arose sooner.

In any case, I appreciate your input and apologize again for the problem.

I opened https://github.com/strongloop/loopback/issues/2196 for this work, and I will address it ASAP.

UniversalOrlando:
@slgoldberg Many thanks for the detailed post! 

For those challenged by this issue using: **loopback-connector-cloudant 1.0.11**, here's the solution that worked for me:  

var filter = { "where": { "name": { "like": "(?i)" + val } } };

UniversalOrlando:
@slgoldberg Many thanks for the detailed post! 

For those challenged by this issue using: **loopback-connector-cloudant 1.0.11**, here's the solution that worked for me:  

var filter = { "where": { "name": { "like": "(?i)" + val } } };

william-screen:
@slgoldberg Many thanks for the detailed post! 

For those challenged by this issue using: **loopback-connector-cloudant 1.0.11**, here's the solution that worked for me:  

var filter = { "where": { "name": { "like": "(?i)" + val } } };

crandmck:
Sadly, this fell through the cracks.... But I went ahead and added a "case-insensitive queries" section to http://loopback.io/doc/en/lb3/AngularJS-JavaScript-SDK.html#read-querying based on @slgoldberg's input above (thanks again).

Better late than never, I guess :-(

william-screen:
Thanks Rand!

________________________________

PR #113, opened 2015-03-19
Return an object when updating models

superkhau:
- The returned object's count property contains the number of models updated

Connected to strongloop/loopback#1167


superkhau:
Closing in favour of #114

PR #114, opened 2015-03-19
Return count when updating or deleting models

superkhau:
- `update/updateAll` returns `info` object
- `destroyAll` return `info` object
- `info` object contains `count` property representing the number of affected items

Connect to strongloop/loopback#1167

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -615,12 +615,12 @@ MongoDB.prototype.destroyAll = function destroyAll(model, where, callback) {
     where = undefined;
   }
   where = self.buildWhere(model, where);
-  this.collection(model).remove(where || {}, function (err, result) {
-    if (self.debug) {
-      debug('destroyAll.callback', model, where, err, result);
-    }
-    var count = result && result.result && result.result.n || 0;
-    callback && callback(err, count);
+  this.collection(model).remove(where || {}, function(err, results) {
+    if (self.debug)
+      debug('destroyAll.callback', model, where, err, results);
+
+    var affectedCount = (results.result && results.result.n) || 0;
+    callback && callback(err, {count: affectedCount});
   });
 };
 
@@ -703,12 +703,12 @@ MongoDB.prototype.update =
     data = self.parseUpdateData(model, data);
 
     this.collection(model).update(where, data, {multi: true, upsert: false},
-      function (err, result) {
-        if (self.debug) {
-          debug('updateAll.callback', model, where, data, err, result);
-        }
-        var count = result && result.result && result.result.n || 0;
-        cb && cb(err, count);
+      function(err, results) {
+        if (self.debug)
+          debug('updateAll.callback', model, where, data, err, results);
+
+        var affectedCount = (results.result && results.result.n) || 0;
+        cb && cb(err, {count: affectedCount});
       });
   };
 
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -433,7 +433,7 @@ describe('mongodb connector', function () {
 
             User.updateAll({age:31},{company:'strongloop.com'},function(err,updatedusers) {
               should.not.exist(err);
-              updatedusers.should.be.equal(2);
+              updatedusers.count.should.be.equal(2);
 
               User.find({where:{age:31}},function(err2,foundusers) {
                 should.not.exist(err2);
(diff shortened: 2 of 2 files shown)


superkhau:
@raymondfeng PTAL

superkhau:
@raymondfeng ^

@bajtos FYI, there are 2 or 3 persisted model suite tests that are still failing on master, but they are not related to the delete/update fixes I'm working on.

lib/mongodb.js
```diff
@@ -615,12 +615,19 @@ MongoDB.prototype.destroyAll = function destroyAll(model, where, callback) {
...
+
+    if (err) return callback(err);
+
+    var info = {
+      count: 0
+    };
+    var affectedCount = results.result.n;
+    if (typeof affectedCount === 'number')
+      info.count = affectedCount;
+    callback && callback(null, info);
```
raymondfeng: It can be simplified as:
```js
var affectedCount = (results.result && results.result.n) || 0;
callback && callback(null, {count: affectedCount});
```

lib/mongodb.js
```diff
@@ -703,12 +710,19 @@ MongoDB.prototype.update =
...
+
+        if (err) return cb(err);
+
+        var info = {
+          count: 0
+        };
+        var affectedCount = results.result.n;
+        if (typeof affectedCount === 'number')
+          info.count = affectedCount;
+        cb(null, info);
```
raymondfeng: Same as above.

raymondfeng:
Added a minor comment. LGTM. Feel free to merge.

lib/mongodb.js
```diff
@@ -703,12 +703,12 @@ MongoDB.prototype.update =
...
-          debug('updateAll.callback', model, where, data, err, result);
-        }
-        var count = result && result.result && result.result.n || 0;
-        cb && cb(err, count);
+      function(err, results) {
+        if (self.debug)
+          debug('updateAll.callback', model, where, data, err, results);
+
+        var affectedCount = (results.result && results.result.n) || 0;
+        cb && cb(err, {count: affectedCount});
```
superkhau: @raymondfeng Fixed. And simplified the `err` by passing it in all the time.

lib/mongodb.js
```diff
@@ -615,12 +615,12 @@ MongoDB.prototype.destroyAll = function destroyAll(model, where, callback) {
...
-      debug('destroyAll.callback', model, where, err, result);
-    }
-    var count = result && result.result && result.result.n || 0;
-    callback && callback(err, count);
+  this.collection(model).remove(where || {}, function(err, results) {
+    if (self.debug)
+      debug('destroyAll.callback', model, where, err, results);
+
+    var affectedCount = (results.result && results.result.n) || 0;
+    callback && callback(err, {count: affectedCount});
```
superkhau: @raymondfeng Fixed. And simplified the `err` by passing it in all the time.

superkhau:
@raymondfeng TY, merging.

lib/mongodb.js
superkhau: @raymondfeng I'm looking back at this now and are we sure this is correct? I think the bracket causes the affectedCount to be true/false or 0. I think we want `result.result && results.result.n ? results.result.n : 0`.

superkhau: Nevermind, it seems the evaluating an object && number leaves you with the rightmost number. However, I will probably still have to fix the fact that we default to zero instead of undefined.

bajtos: > However, I will probably still have to fix the fact that we default to zero instead of undefined.

Yes please. `result.result ? result.result.n : undefined` may be good enough.

test/mongodb.test.js
```diff
@@ -462,7 +462,7 @@ describe('mongodb connector', function () {
 
             User.updateAll({}, {age: 40, '$set': {age: 39}},function(err,updatedusers) {
               should.not.exist(err);
-              updatedusers.should.be.equal(3);
+              updatedusers.count.should.be.equal(3);
```
bajtos: Consider using `updatedusers.should.have.property(count, 3)` instead, it will give you nicer failure messages. Another nice-to-have change: rename `updatedusers` to `info`. This comment applies to all tests you are changing in this patch, feel free to ignore it though.

PR #116, opened 2015-03-24
Pass results to DAO save and upsert callback

rmg:
Added to support ctx.isNewInstance in strongloop/loopback-datasource-juggler#515.


bajtos:
Closing in favour of #121

PR #118, opened 2015-03-26
Default to `undefined` instead of `0`

superkhau:
On `updateAll` or `destroyAll`, default to `undefined` instead of `0`.

Connect to strongloop/loopback#1167


superkhau:
@raymondfeng PTAL

raymondfeng:
Lgtm

bajtos:
:+1:

PR #119, opened 2015-03-30
Cleanup for returning count on update/delete

superkhau:
@bajtos PTAL, adding the code review updates based on your comments from strongloop/loopback#1167


bajtos:
Cool, :shipit:

PR #124, opened 2015-04-09
Feature/benchmarks

superkhau:
Spike to see what needs to be done regarding a basic benchmark for one connector.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,3 +11,7 @@ test-verbose:
 testing:
 	$(TESTER) $(OPTS) --watch $(TESTS)
 .PHONY: test docs
+
+b benchmark benchmarks:
+	@node benchmarks >> $(CURDIR)/benchmarks/results.md \
+		&& echo 'Done. See ./benchmarks/results.md'
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -1,0 +1,77 @@
+var DataSource = require('loopback-datasource-juggler').DataSource;
+var connector = require('..');
+var Benchmark = require('benchmark');
+
+var ds = new DataSource(connector, {
+  host: process.env.LB_HOST || '127.0.0.1',
+  port: process.env.LB_PORT || 27017,
+  database: process.env.LB_DB || 'strongloop'
+});
+var Todo = ds.define('Todo', {
+  content: {type: String}
+});
+
+// not critical for MongoDB, but may uncover inefficiencies in SQL connectors
+// https://github.com/strongloop/loopback-connector-mongodb/pull/124/files#r28435614
+var uniqVal = 0;
+
+function resetTestState() {
+  uniqVal = 0;
+  Todo.destroyAll();
+}
+
+var suite = new Benchmark.Suite;
+suite
+  .on('start', function() {
+    console.log('#', new Date());
+  })
+  .add('create', {
+    defer: true,
+    fn: function(deferred) {
+      Todo.create({content: 'Buy eggs, ' + (uniqVal++)}, function() {
+        deferred.resolve();
+      });
+    },
+    onComplete: resetTestState
+  })
+  .add('find', {
+    defer: true,
+    fn: function(deferred) {
+      Todo.find(function() {
+        deferred.resolve();
+      });
+    },
+    onStart: function() {
+      Todo.create([
+        {content: 'Buy eggs'},
+        {content: 'Buy milk'},
+        {content: 'Buy cheese'}
+      ]);
+    },
+    onComplete: resetTestState
+  })
+  .add('find with a simple filter', {
+    defer: true,
+    fn: function(deferred) {
+      Todo.find({where: {content: 'Buy milk'}}, function() {
+        deferred.resolve();
+      });
+    },
+    onStart: function() {
+      Todo.create([
+        {content: 'Buy eggs'},
+        {content: 'Buy milk'},
+        {content: 'Buy cheese'}
+      ]);
+    },
+    onComplete: resetTestState
+  })
+  .on('cycle', function(event) {
+    console.log(String(event.target));
+  })
+  .on('complete', function() {
+    console.log();
+    Todo.destroyAll();
+    process.exit();
+  })
+  .run({async: true});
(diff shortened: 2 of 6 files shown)


superkhau:
Connect to strongloop/loopback#1291

superkhau:
@raymondfeng This is a simple spike using `memwatch` to detect leaks when creating models. No leaks are detected for this particular method. Anyways, let me know if you have any opinions on this and whether the approach is sufficient.

benchmarks/index.js
```diff
@@ -0,0 +1,33 @@
...
+  .add('Todo.find', function() {
+    Todo.find();
+  })
+  .on('cycle', function(event) {
+    console.log(String(event.target));
+  })
+  .on('complete', function(event) {
+    console.log();
+    Todo.destroyAll();
+    process.exit();
```
superkhau: @raymondfeng The node app stays open when done. I think this is because the datasource still has a connection open to the database. Is there a way to force a disconnect manually (ie. `ds.connector.disconnect`)?

superkhau:
@raymondfeng Please review.

/cc @bajtos @ritch

benchmarks/index.js
raymondfeng: You can call `ds.disconnect(cb);`

benchmarks/results.md
```diff
@@ -0,0 +1,4 @@
+# Tue Apr 14 2015 09:34:23 GMT-0700 (PDT)
+Todo.create x 110,662 ops/sec ±13.34% (67 runs sampled)
+Todo.find x 76,328 ops/sec ±49.29% (51 runs sampled)
+
```
raymondfeng: Do you want to add this file to .gitignore?

superkhau: I think we should leave it in VC so we can keep track of the results over time. Like next month when we run it, we can compare it to the this month.

benchmarks/index.js
superkhau: Hm..doesn't work. I'll have to force it through `process.exit()` for now.

raymondfeng:
Where is the usage of `memwatch`? I couldn't find it.

Makefile
```diff
@@ -11,3 +11,7 @@ test-verbose:
 testing:
 	$(TESTER) $(OPTS) --watch $(TESTS)
 .PHONY: test docs
+
+b benchmarks:
```
raymondfeng: Is it a typo?

superkhau:
`memwatch` is part of the other PR https://github.com/strongloop/loopback-connector-mongodb/pull/126/files#diff-cf7a6402d000b39b289216c3615be4c2R1 as it's only used for leak detection and not benchmarking.

Makefile
superkhau: This is not a typo, Makefiles allow multi-target, so instead of typing `make benchmarks` all the time, I like to alias `b` so you can type `make b` or `make benchmarks`. Seeing this now, I should also add `benchmark` in case the user forgets to type the `s` in `benchmarks`

benchmarks/index.js
```diff
@@ -0,0 +1,34 @@
+var DataSource = require('loopback-datasource-juggler').DataSource;
+var connector = require('..');
+var Benchmark = require('benchmark');
+
+var ds = new DataSource(connector, {
```
superkhau: @raymondfeng Is this the right way to initialize a datasource manually? I'm starting to think this is why it doesn't disconnect.

Maybe it should be:
```
var ds = new DataSource('mongodb', opts)
var connector = new Connector(ds, cb);
```

I noticed because looking https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L69-L71, it seems like we only connect if a callback is provided.

superkhau:
@raymondfeng I updated the benchmark code to use deferred, it works properly now (actually creating items/connecting to db). Please review again.

benchmarks/index.js
```diff
@@ -0,0 +1,43 @@
...
+      });
+    }
+  })
+  .on('cycle', function(event) {
+    console.log(String(event.target));
+  })
+  .on('complete', function() {
+    console.log();
+    Todo.destroyAll();
+    process.exit();
```
superkhau: @raymondfeng This lib [does not support async setup/teardown](https://github.com/bestiejs/benchmark.js/issues/70). So we have two choices:

1. `process.exit()` to force the script to quit (items in the db are cleaning up properly since its fire and forget anyways)
2. Write a sync version of `destroyAll` in the connector and use that instead.

benchmarks/results.md
```diff
@@ -0,0 +1,4 @@
+# Tue Apr 14 2015 12:36:47 GMT-0700 (PDT)
+Todo.create x 233 ops/sec ±100.80% (73 runs sampled)
```
superkhau: @raymondfeng I have no idea why this is `±100.80%` (relative margin or error). Does this mean it can be up to 100% wrong? lol

benchmarks/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+var suite = new Benchmark.Suite;
+suite
+  .on('start', function() {
+    console.log('#', new Date());
+  })
+  .add('Todo.create', {
+    defer: true,
+    fn: function(deferred) {
+      Todo.create({content: 'Buy eggs'}, function() {
```
bajtos: Please use a unique value here, e.g. `'Buy eggs' + (tid++)` where `var tid = 0` is set somewhere outside of `fn`. While it's probably not critical for MongoDB, it can uncover inefficiencies in SQL connectors.

benchmarks/index.js
```diff
@@ -0,0 +1,43 @@
...
+        deferred.resolve();
+      });
+    }
+  })
+  .add('Todo.find', {
+    defer: true,
+    fn: function(deferred) {
+      Todo.find(function() {
```
bajtos: Please add another benchmark that builds at least a simple filter (query).

benchmarks/index.js
superkhau: I'm guessing `tid` is `todoId`?

benchmarks/index.js
superkhau: Done.

superkhau:
@bajtos TFL, updated with your CR comments.

superkhau:
@raymondfeng You asked how `benchmark.js` works the other day, see http://stackoverflow.com/questions/4986245/how-does-jsperf-work

bajtos:
As far as I am concerned, this LGTM. Please wait for @raymondfeng's approval before merging.

raymondfeng:
LGTM

PR #126, opened 2015-04-14
Feature/leak detection

superkhau:
- Add test to ensure leak detector actually detects leaks
- Add leak detection tests for `find` operation
- Add leak detection tests for `create` operation
- Add instruction to README.md for running leak detection tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "main": "index.js",
   "scripts": {
     "benchmarks": "make benchmarks",
+    "leak-detection": "make leak-detection",
     "test": "make test"
   },
   "dependencies": {
@@ -24,10 +25,12 @@
     "benchmark": "^1.0.0",
     "bluebird": "^2.9.12",
     "loopback-datasource-juggler": "^2.15.0",
+    "memwatch": "^0.2.2",
     "mocha": "^2.1.0",
     "rc": "^1.0.0",
     "semver": "^4.2.0",
-    "should": "^5.0.0"
+    "should": "^5.0.0",
+    "sinon": "^1.14.1"
   },
   "repository": {
     "type": "git",
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,15 @@ testing:
 	$(TESTER) $(OPTS) --watch $(TESTS)
 .PHONY: test docs
 
+.PHONY: b benchmark benchmarks
 b benchmark benchmarks:
 	@node benchmarks >> $(CURDIR)/benchmarks/results.md \
 		&& echo 'Done. See ./benchmarks/results.md'
+
+.PHONY: l ld leak leak-detection
+l ld leak leak-detection:
+	@ITERATIONS=$(ITERATIONS) $(TESTER) leak-detection \
+		--recursive \
+		--reporter spec \
+		--require $(CURDIR)/leak-detection/globals \
+		--timeout 60000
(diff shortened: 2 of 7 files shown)


superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@raymondfeng Review please.

/cc @bajtos @ritch

package.json
```diff
@@ -21,11 +22,14 @@
...
     "bluebird": "^2.9.12",
+    "chai": "^2.2.0",
     "loopback-datasource-juggler": "^2.15.0",
+    "memwatch": "^0.2.2",
     "mocha": "^2.1.0",
     "rc": "^1.0.0",
     "semver": "^4.2.0",
-    "should": "^5.0.0"
+    "should": "^5.0.0",
+    "sinon": "^1.14.1"
```
bajtos: Since `sinon` is used in one place only, please consider rewriting that place in such way that we are not adding another dev dependency.

package.json
```diff
@@ -21,11 +22,14 @@
   },
   "devDependencies": {
     "bluebird": "^2.9.12",
+    "chai": "^2.2.0",
```
bajtos: Please keep the test code consistent and use the `should` module in the memory leak tests too.

package.json
bajtos: Oh, it's used in both test files, never mind then. I am deleting the previous comment.

bajtos:
Is there a way how to run all existing tests with memwatch enabled? That way we don't have to write specific memwatch tests for each connector method.

Perhaps it's enough to add the following global before hook to "test/init.js"?

```js
before(setupMemwatch() {
  var memwatch = require('memwatch');
  memwatch.on('leak', function() {
    throw new Error('MEMORY LEAK DETECTED');
  });
});
```

Note that Mocha intercepts unhandled errors and marks the currently running test as failed in such case.

superkhau:
>Is there a way how to run all existing tests with memwatch enabled? That way we don't have to write specific memwatch tests for each connector method.

What you said would work, but it would only do it for 1 iteration (which is obviously useless). I am not aware of a way to run multiple iterations without rewriting the existing test code.

package.json
superkhau: Done.

superkhau: Actually, question for you. Is it considered best practice to not introduce new dependencies because we already established an assertion library to use? Does it make more sense in this case to use the lib already built in and if we decide to move onto something like chai, to do a full refactor then? Trying to understand why I should use `chai` as it allows for custom error messages (and I feel it's better than should.js).

bajtos:
> What you said would work, but it would only do it for 1 iteration (which is obviously useless).

Oh, I see, it's because "A leak event will be emitted when your heap usage has increased for five consecutive garbage collections". I didn't know how memwatch works under the hood, now I understand why my idea won't work.

Consider this thread as resolved then.

superkhau:
>"A leak event will be emitted when your heap usage has increased for five consecutive garbage collections"

Yes, this is the key. I actually like your idea better (because having the tests themselves auto-check for leak detection would be sweet since we wouldn't have to write individual tests for each function).

As a discussion topic, I am more interested in how we can reduce the number of functions to test. I see this from either a unit test level or integration test level. I mean technically if the individual units do not leak, the components together technically shouldn't leak (but this might be a weak assumption). However, for robustness, its probably best to check at both levels (I see unit level as critical and integration level as nice-to-have).

raymondfeng:
LGTM. Can you rebase?

superkhau:
Rebased and merged. TY

PR #127, opened 2015-04-16
add documentation on enabling extended operators

Neil-UWA:
add documentation on how to use mongo built-in operators


slnode:

PR #130, opened 2015-05-15
Set batchSize to the same as the limit size to ensure correct number of results is returned

### LINKED ISSUE #129: Limiting the number of query results incorrectly applied due to batchSize
When using the mongo connector, applying a limit to a query does not limit the results properly. Actually, the number of results returned corresponds to the batchSize that is set for Mongo (1000 in my case). Query limits work correctly up to the point where the batchSize is reached.

For example:

```
curl -s "http://127.0.0.1:3000/api/v1/Events?filter\[limit\]=100" | jq '. | length' # returns 100 results
curl -s "http://127.0.0.1:3000/api/v1/Events?filter\[limit\]=500" | jq '. | length' # returns 500 results
curl -s "http://127.0.0.1:3000/api/v1/Events?filter\[limit\]=1000" | jq '. | length' # returns 1000 results
curl -s "http://127.0.0.1:3000/api/v1/Events?filter\[limit\]=1200" | jq '. | length' # returns 2000 results
curl -s "http://127.0.0.1:3000/api/v1/Events?filter\[limit\]=2500" | jq '. | length' # returns 3000 results
```

and so on. (I'm using http://stedolan.github.io/jq/ to count the returned results easily).

mrfelton:
Set batchSize to the same as the limit size to ensure correct number of results is returned [fixes #129]

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -591,6 +591,7 @@ MongoDB.prototype.all = function all(model, filter, callback) {
 
   if (filter.limit) {
     cursor.limit(filter.limit);
+    cursor.batchSize(filter.limit);
   }
   if (filter.skip) {
     cursor.skip(filter.skip);
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -591,36 +591,36 @@ MongoDB.prototype.all = function all(model, filter, callback) {
 
   if (filter.limit) {
     cursor.limit(filter.limit);
-    cursor.batchSize(filter.limit);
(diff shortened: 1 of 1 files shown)


slnode:

raymondfeng:
@mrfelton Thanks for the patch. 

1. Please sign the CLA at https://cla.strongloop.com/agreements/strongloop/loopback-connector-mongodb
2. Can you add some comments about the change?
3. Is there a way to add a unit test?

PR #138, opened 2015-06-01
Clarkorz feature/optmized find or create

paran01d:
Hi I found issue #92 because I was looking for exactly this in the mongodb connector, looks like it got left behind because of an upstream dependency, so I thought Id check it out, needed to update a conflict in the package.json to get the tests to work with v1.9.2 but seems working.


slnode:

PR #143, opened 2015-06-07
Replaced ensureIndex() with createIndex()

uzyn:
as `ensureIndex()` is now deprecated.

Reference: http://docs.mongodb.org/manual/reference/method/db.collection.ensureIndex/


slnode:

PR #150, opened 2015-06-17
Race condition listening to fullsetup event from an db object in a callback

jsheely:
There is a race condition listening to the db.on('fullsetup') event from the mongodb db module. 

Because the mongodb module does not fire events on process.nextTick() then the callback may get fired after the event has already run which yields a race condition where db.authenticate never gets fired.

This will cause situation where the juggler will see a timeout in connecting to the database simply because the callback never gets called.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -149,8 +149,7 @@ function connectToMongoDB(url, options, callback) {
       if (err) return callback(err);
 
       if (db) {
-        db.on('fullsetup', function() {
-          db.authenticate(user, password, options, function(err, ok) {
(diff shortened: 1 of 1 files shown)


slnode:

raymondfeng:
@jsheely Thank you for the patch. The workaround was introduced for https://jira.mongodb.org/browse/NODE-478. Now the bug is fixed in mongodb driver, we'll revert our workaround as soon as mongodb@2.0.34 is published.

raymondfeng:
Upgraded to mongodb@2.0.35 and reverted my workaround to https://jira.mongodb.org/browse/NODE-478. The issue should be fixed now.

PR #158, opened 2015-07-06
Added a setting to enable optimsied findOrCreate method 

paran01d:
so that connector continues to work with mongodb < 2.6. Not sure if this is acceptable but wanted to get some feedback.


slnode:

PR #159, opened 2015-07-14
Update benchmarks

superkhau:


superkhau:
@raymondfeng Merging without review.

Issue #163, opened 2015-07-22
Honor Model.definition.settings.allowExtendedOperators


fabien:
See: https://github.com/strongloop/loopback-datasource-juggler/pull/666#issuecomment-123754151

raymondfeng:
@fabien Do you want to submit a patch? It is very straightforward. :-)

fabien:
@raymondfeng patch is ready #164

PR #164, opened 2015-07-23
Enable Model.settings.mongodb.allowExtendedOperators

fabien:
Implements #163 as follows:

- when enabled on the connector (`=== true`), all models will allow extended operators
- when disabled on the connector (`=== false`), models will not allow extended operators at all, ever
- when enabled on the connector, disabling the model setting (`=== false`) will prevent extended operators for that single model
- when enabled on the connector, enabling the model setting has no effect (it remains enabled)
- when not explicitly set on the connector, use the model's setting appropriately (`true / false`)

/cc @raymondfeng

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -383,13 +383,16 @@ MongoDB.prototype.find = function find(model, id, options, callback) {
  * @param data
  * @returns {*}
  */
-MongoDB.prototype.parseUpdateData = function(model, data) {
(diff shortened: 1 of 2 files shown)


fabien:
In addition, you can pass `{ allowExtendedOperators: true/false }` with your method call to explicitly enable or disable extended operators - options takes precedence over any other settings.

```
User.updateAll({name: 'Al'}, {'$rename': {name: 'firstname'}}, { allowExtendedOperators: true }, cb);
```

PR #166, opened 2015-08-04
Add regexp operator support

superkhau:
Connect to strongloop/loopback-datasource-juggler#552


superkhau:
@raymondfeng PTAL

raymondfeng:
LGTM

PR #167, opened 2015-08-05
support to buffer for embedded models

FelipeAguayo:
It works for me.

I found that support for embedded models was not implemented. According to the implemented, it is supported buffer type for 'embedsMany' and 'embedsOne'.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -170,7 +170,6 @@ MongoDB.prototype.fromDatabase = function (model, data) {
   if (!data) {
     return null;
   }
-  console.log(data)
   if(typeof model === 'string'){
     model = this._models[model];
     props = model.properties;
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -165,7 +165,7 @@ MongoDB.prototype.collection = function (model) {
  */
 MongoDB.prototype.fromDatabase = function (model, data) {
   var self = this
-    , props, objModel
+    , props
   ;
   if (!data) {
     return null;
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -167,7 +167,7 @@ MongoDB.prototype.fromDatabase = function (model, data) {
   var self = this;
   var props;
   var settings;
-  var relations = [];
+  var relations = {};
   if (!data) {
     return null;
   }
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -166,7 +166,7 @@ MongoDB.prototype.collection = function (model) {
 MongoDB.prototype.fromDatabase = function (model, data) {
   var self = this;
   var props;
-  var settings;
+  var modelClass;
   var relations = {};
   if (!data) {
     return null;
@@ -175,12 +175,9 @@ MongoDB.prototype.fromDatabase = function (model, data) {
     model = this._models[model];
   }
   props = model.properties;
-  settings = model.settings;
-  if(settings){
-    for(var k in settings.relations){
-      if(settings.relations[k].property){
-        relations[settings.relations[k].property] = settings.relations[k];
-      }
+  for(var k in model.model.relations){
+    if(model.model.relations[k].keyFrom){
+      relations[model.model.relations[k].keyFrom] = model.model.relations[k];
     }
   }
   for (var p in props) {
@@ -197,19 +194,21 @@ MongoDB.prototype.fromDatabase = function (model, data) {
       }
     } // check if prop is a relation EmbedsMany
     else if(prop && relations[p] && relations[p].type === 'embedsMany'){
+      modelClass = relations[p].modelTo;
       for(var i = 0; i < data[p].length; i++){
         data[p][i] = self.fromDatabase({
-          model: prop.type[0],
-          properties: prop.type[0].definition.properties,
-          settings: prop.type[0].definition.settings
+          model: modelClass,
+          properties: modelClass.definition.properties,
+          settings: modelClass.definition.settings
         },data[p][i]);
       }
     } // check if prop is a relation EmbedsOne
     else if(prop && relations[p] && relations[p].type === 'embedsOne'){
+      modelClass = relations[p].modelTo;
       data[p] = self.fromDatabase({
-        model: prop.type,
-        properties: prop.type.definition.properties,
-        settings: prop.type.definition.settings
+        model: modelClass,
+        properties: modelClass.definition.properties,
+        settings: modelClass.definition.settings
       },data[p]);
     }
   }
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -165,7 +165,7 @@ MongoDB.prototype.collection = function (model) {
  */
 MongoDB.prototype.fromDatabase = function (model, data) {
   var self = this
-    , props, settings, objModel
+    , props, objModel
   ;
   if (!data) {
     return null;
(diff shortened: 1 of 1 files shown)


lib/mongodb.js
```diff
@@ -180,6 +188,21 @@ MongoDB.prototype.fromDatabase = function (model, data) {
         // Convert the Binary into String
         data[p] = data[p].toString();
       }
+    }  // check if prop contains EmbedsMany
+    else if(prop && prop.type.constructor === Array &&
```
raymondfeng: There is a better way to check relations. You can access modelClass.definition.relations.

lib/mongodb.js
```diff
@@ -164,10 +164,18 @@ MongoDB.prototype.collection = function (model) {
  * @param {Object} data The data from DB
  */
 MongoDB.prototype.fromDatabase = function (model, data) {
+  var self = this
+    , props
+  ;
```
raymondfeng: We use the following style:
```
var self = this;
var props;
```

raymondfeng:
Can you elaborate what problem this PR tries to address?

lib/mongodb.js
FelipeAguayo: it's done.

FelipeAguayo:
mongodb data:
```
{
    "_id" : ObjectId("55c2781534db8a18135c607c"),
    "name" : "model example",
    "img" : [ 
        {
            "buffer" : { "$binary" : "binary code", "$type" : "00" },
            "name" : "2333461d13722e354a5342ab.bmp",
            "id" : "2333461d13722e354a5342ab",
        }
    ]
}
```
parent model:
```
{
  "name": "ModelExample",
  "plural": "ModelExamples",
  "base": "PersistedModel",
  "idInjection": true,
  "strict" : true,
  "properties": {
    "name":"string"
  },
  "validations": [],
  "hidden": [],
  "protected": [],
  "relations": {
    "images":{
      "type": "embedsMany",
      "model": "Image",
      "property": "img",
      "options": {
        "validate": true,
        "forceId": false
      }
    }
  },
  "acls": [],
  "methods": []
}
```
child model:
```
{
  "name": "Image",
  "plural": "Image",
  "base": "PersistedModel",
  "idInjection": true,
  "strict" : true,
  "properties": {
    "name":"string",
    "buffer": {
      "type": "Buffer",
      "required": true
    }
  },
  "hidden": [],
  "protected":[],
  "validations": [],
  "relations": {},
  "acls": [],
  "methods": []
}
```
based on the above, when the data is loaded from mongodb, the "buffer" of the child model is not charged.

lib/mongodb.js
FelipeAguayo: the problem is that the property name is not the relation's name. 
but in the previous commit ( c432359cbc71cb6db9b6f24049660adc73f07a11 ) I proposed the following:

```javascript
MongoDB.prototype.fromDatabase = function (model, data) {
  var self = this;
  var props;
  var settings;
  var objModel;
  ;
  if (!data) {
    return null;
  }
  if(typeof model === 'string'){
    model = this._models[model];
    props = model.properties;
    settings = model.settings;
  }else{
    props = model.properties;
    settings = model.settings;
  }
  for (var p in props) {
    var prop = props[p];
    if (prop && prop.type === Buffer) {
      if(data[p] instanceof mongodb.Binary) {
        // Convert the Binary into Buffer
        data[p] = data[p].read(0, data[p].length());
      }
    } else if(prop && prop.type === String) {
      if(data[p] instanceof mongodb.Binary) {
        data[p] = data[p].toString();
      }
    } // check if prop contains EmbedsMany
    else if(prop && prop.type.constructor === Array &&
        prop.type[0].toString() === model.model.toString()){
      for(var k in settings.relations){
        if(settings.relations[k].property === p){
          for(var i = 0; i < data[p].length; i++){
            data[p][i] = self.fromDatabase({
              model: prop.type[0],
              properties: prop.type[0].definition.properties,
              settings: prop.type[0].definition.settings
            },data[p][i]);
          }
        }
      }
    } // check if prop contains EmbedsOne
    else if(prop && prop.type.toString() === model.model.toString()){
      for(var k in settings.relations){
        if(settings.relations[k].property === p){
          data[p] = self.fromDatabase({
            model: prop.type,
            properties: prop.type.definition.properties,
            settings: prop.type.definition.settings
          },data[p]);
        }
      }
    }
  }
  return data;
};
```

its works, but there are many for.

FelipeAguayo: @raymondfeng : I implemented the method ( 3d5b38292aa42dfed6a4e5f5bca68cfa53329ea6 ), now this check if the property corresponds to a relation in the model:

```javascript
/*!
 * Convert the data from database to JSON
 *
 * @param {String|Object} model The model name or the model class
 * @param {Object} data The data from DB
 */
MongoDB.prototype.fromDatabase = function (model, data) {
  var self = this;
  var props;
  var settings;
  var relations = [];
  if (!data) {
    return null;
  }
  if(typeof model === 'string'){
    model = this._models[model];
  }
  props = model.properties;
  settings = model.settings;
  if(settings){
    for(var k in settings.relations){
      if(settings.relations[k].property){
        relations[settings.relations[k].property] = settings.relations[k];
      }
    }
  }
  for (var p in props) {
    var prop = props[p];
    if (prop && prop.type === Buffer) {
      if(data[p] instanceof mongodb.Binary) {
        // Convert the Binary into Buffer
        data[p] = data[p].read(0, data[p].length());
      }
    } else if(prop && prop.type === String) {
      if(data[p] instanceof mongodb.Binary) {
        data[p] = data[p].toString();
      }
    } // check if prop is a relation EmbedsMany
    else if(prop && relations[p] && relations[p].type === 'embedsMany'){
      for(var i = 0; i < data[p].length; i++){
        data[p][i] = self.fromDatabase({
          model: prop.type[0],
          properties: prop.type[0].definition.properties,
          settings: prop.type[0].definition.settings
        },data[p][i]);
      }
    } // check if prop is a relation EmbedsOne
    else if(prop && relations[p] && relations[p].type === 'embedsOne'){
      data[p] = self.fromDatabase({
        model: prop.type,
        properties: prop.type.definition.properties,
        settings: prop.type.definition.settings
      },data[p]);
    }
  }
  return data;
};
```

lib/mongodb.js
```diff
@@ -160,14 +160,29 @@ MongoDB.prototype.collection = function (model) {
...
  *
- * @param {String} model The model name
+ * @param {String|Object} model The model name or the model class
  * @param {Object} data The data from DB
  */
 MongoDB.prototype.fromDatabase = function (model, data) {
+  var self = this;
+  var props;
+  var settings;
+  var relations = [];
```
raymondfeng: It should be `{}`.

lib/mongodb.js
```diff
@@ -160,14 +160,29 @@ MongoDB.prototype.collection = function (model) {
...
+  }
+  props = model.properties;
+  settings = model.settings;
+  if(settings){
+    for(var k in settings.relations){
+      if(settings.relations[k].property){
+        relations[settings.relations[k].property] = settings.relations[k];
+      }
+    }
+  }
```
raymondfeng: relations can be accessed as `model.relations`. No need to introspect `settings` which only contains relations from the json file.

lib/mongodb.js
FelipeAguayo: you are right

lib/mongodb.js
FelipeAguayo: I had to do this because I have to check if the property delivered by mongodb is defined as that property in the relation, because it is stored as property and not as a relation:

```json
"images":{
      "type": "embedsMany",
      "model": "Image",
      "property": "img",
      "options": {
        "validate": true,
        "forceId": false
      }
```

FelipeAguayo: @raymondfeng : please check ( 9b83047d5db42ec5da15472f81c262f843a153f3 )

FelipeAguayo: @raymondfeng : implemented with access to model.relations. please check ( 9b83047 )

PR #169, opened 2015-08-12
Relax id requirement for CRUD operations

superkhau:
Connect to strongloop/loopback-datasource-juggler#670

### DIFF
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -1481,7 +1481,7 @@ describe('mongodb connector', function () {
 
   context('id', function() {
     var Todo = getDataSource().define('Todo', {
-      content: {type: String}
+      content: {type: String, id: false}
     });
 
     before(function deleteExistingTestFixtures(done) {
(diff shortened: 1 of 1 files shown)


test/mongodb.test.js
```diff
@@ -1479,6 +1479,34 @@ describe('mongodb connector', function () {
     });
   });
 
+  context('id', function() {
+    var Todo = getDataSource().define('Todo', {
+      content: {type: String}
+    });
```
superkhau: @raymondfeng Does this make sense? Define a model without an id

test/mongodb.test.js
```diff
@@ -1479,6 +1479,34 @@ describe('mongodb connector', function () {
...
+        {content: 'Buy sausages'}
+      ], done);
+    });
+    after(function deleteTestFixtures(done) {
+      Todo.destroyAll(done);
+    });
+
+    it.only('should not be required when when performing a find query',
+        function(done) {
+      Todo.find(function(err, posts) {
```
superkhau: Then query it to make sure it executes without error?

Issue #52, opened 2014-09-11
Why converting string to ObjectId in 'inq' filter?


coodoo:
I get the idea behind doing so is to allow querying by _id, but the problem it's causing is any string of length 12 or 24 will be converted too, even worser if those strings matches ObjectID format, they will be converted to ObjectID type when they should remain String type.

In short, this kind of "forced" conversion is kind of a surprise and smell like a hack?

``` js
} else if (spec === 'inq') {
        query[k] = { $in: cond.map(function (x) {
          if ('string' !== typeof x) return x;
          return ObjectID(x);
        })};
      } 
```

See full code here:
http://goo.gl/Xwhlj4

raymondfeng:
@SSchaaf Can you submit your fix as a patch?

terrence0488:
@raymondfeng hi Raymond, is this going to be fixed? Right now, if I would like to do a inq search to a non id field, how can i do it? Im trying to avoid changing code in the mongo-connector.

wrensburg:
:+1: Having to hack the framework to achieve this!

chrisamoore:
:+1: Def need this. /cc @ShubhraKar @dashby3000

ekussberg:
@raymondfeng @bajtos is this problem patched/fixed? We had several times already this issue and needed to change the type of ID to string from ObjectId as INQ just converts the String always to ObjectId and MongoDB return zero results. Could you provide a fix or solution how to make INQ with simple strings? I tried already this
```
"deliveryTourId": {
      "type": "ObjectId",
}
```
It also doesnt work. If i put string, then it returns 0 as string != objectId.

raymondfeng:
Can you try with https://github.com/strongloop/loopback-connector-mongodb/pull/174?

totalard:
Not fixed Yet

shallker:
Still not fixed!!!!

shallker:
if the string length is 24, it will be always treated as ObjectId automatically, what is the purpose of this design?

bajtos:
I am reopening the issue then. @strongloop/sq-lb-apex could you PTAL?

vinzentt:
This bug also breaks RoleMapping built-in model requests. 
It has a string field called principalId, so if you try to find the roles for a user, you cannot, you have to access MongoDb native connector via `MyModel.getDataSource().connector.collection(MyModel.modelName)`.

anjanarana:
Hi Guys,

[...]

PR #174, opened 2015-08-25
Enhance coercion of ids with inq/nin operators

raymondfeng:
Fix https://github.com/strongloop/loopback-connector-mongodb/issues/52

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -87,8 +87,11 @@ function MongoDB(settings, dataSource) {
   }
 
   this.dataSource = dataSource;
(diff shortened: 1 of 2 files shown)


cthanhnguyen:
this is not fix now, Why you guys abandon this fixed

superkhau:
@cthanhnguyen Please create your own issue describing your situation and a link to a repo that we can clone to reproduce the issue. We generally do not like to comment on merged/closed issues.

shallker:
if the string length is 24, it will be always treated as ObjectId automatically, do a test BYS

Issue #147, opened 2015-06-12
Default sort order overwrite the $nearSphere criteria


kwiky:
When i use the $nearSphere criteria, the default order on id overwrite the location order :(

Here : 
https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L653

mazesoul:
+1

jskrzypek:
+1

ssh24:
Related to https://github.com/strongloop/loopback-connector-mongodb/issues/55. Closing this. Feel free to reopen if needed.

Issue #55, opened 2014-09-30
Default sort order may impact performance on some queries


mihail-stanchev:
If a sort order is not specified a default one by _id is used. This may impact the performance of some queries where a sort is neither needed nor specified upfront. For example, using model.findOne results in the MongoDB connector calling db.collection.find(query).sort({_id:1}).limit(1)
If the sort part is skipped the result is returned almost immediately using the appropriate indices. However, having the sort in there causes the entire query to slow down. In my case there is only one record matching the given criteria and the sort is not needed.

Would you consider making the sort order explicit?

raymondfeng:
Maybe we can make it controllable via a model/method level configuration property.

mazesoul:
+1

Ijmir:
+1

thaiat:
+1

SynerG:
+1 Please, fix this. Even if you create a specific index for your query, MongoDb cannot use the index because Loopback requests to order by `_id`, and the field is not included in the index. MongoDB docs explain it: https://docs.mongodb.org/manual/core/index-intersection/#index-intersection-and-sort

sebastianhaas:
+1

simonecogno:
+1

0101adm:
+1

bidos:
+1

sebastianhaas:
This is still relevant to me!

sebastianhaas:
Don't!

HugoPoi:
I loose one day because of this default sort operation on "_id". THE DEFAULT BEHAVIOR FOR MONGODB MUST BE NO SORT when no sort operation is specify !
Please merge  #177 !

HugoPoi:
UP This issue is still relevant for 1.18.x and 3.x.

See 
https://github.com/strongloop/loopback-connector-mongodb/blob/c0c75f915ea54079e66c819e28036b7c7bf6f7d6/lib/mongodb.js#L729-L734
And
https://github.com/strongloop/loopback-connector-mongodb/blob/0041f217019c84983ba3eec39c0ac993807289fc/lib/mongodb.js#L849-L854

Can we re-open and merge https://github.com/strongloop/loopback-connector-mongodb/pull/177 ?

sebastianhaas:
Don't

dhmlau:
[...]

PR #177, opened 2015-09-16
Add the option to bypass the default sort order by using option.useDefaultSort: false

mazesoul:
This is useful when using operators that provide their own sort. For instance `nearSphere` returns the results from lowest to greatest distance.

Addresses the following issues #147, #55 

```
Model.find({}, {useDefaultSort:false});
```

The bypass will only trigger if there is no `order` in the filter and the value is explicitely set to `false`

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -682,17 +682,20 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
   }
 
   function processResponse(err, cursor) {
+    var useDefaultSort = options.useDefaultSort !== false;
+
     if (err) {
       return callback(err);
     }
-    var order = {};
-    if (!filter.order) {
+    var order = null;
+    if (!filter.order && useDefaultSort) {
       var idNames = self.idNames(model);
       if (idNames && idNames.length) {
         filter.order = idNames;
       }
     }
     if (filter.order) {
+      order = {};
       var keys = filter.order;
       if (typeof keys === 'string') {
         keys = keys.split(',');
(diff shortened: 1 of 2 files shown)


slnode:

rmg:
@slnode ok to test

mazesoul:
@rmg Could you help me figure out the failures?

So far, it seems that most of it are build system related:

- sh: mocha: command not found
- Package name: loopback-getting-started does not match $PACKAGE_NAME: loopback-getting-started-intermediate

Thanks

slnode:

raymondfeng:
@mazesoul Don't worry about such failures which are not directly related to your PR. The build tries to make sure all downstream modules are tested with your PR.

mazesoul:
@raymondfeng tx. It looked unrelated to the change but I wanted to make sure

raymondfeng:
@mazesoul Two thoughts around the proposal:

1. Should we provide a configuration option at app/model level to control if order by id should be enforced?
2. We should support the same option for all connectors.

Are you willing to help us complete the solution?

mazesoul:
@raymondfeng 

1. I think the default ordering by id is ok and useful in almost all cases and wouldn't expand the option to all cases unless the performance issue is a concern. #55 
2. I haven't looked at other connectors but I guess it makes sense to have it applicable everywhere

As for helping with other connector, I can definitely look at it. I'm not currently setup to test on all the supported dbs but it could be done :)

mazesoul:
@raymondfeng Any suggestion concerning the follow up? Should I start on the other connector before this can be merge. Would you prefer I add support for the option at all levels of the models?

rmg:
@slnode test please

ssh24:
@mazesoul Are you willing to pick up the work for this PR again? Let's get started by rebasing the branch and fixing the merge conflicts.

I also agree with you on this
> I think the default ordering by id is ok and useful in almost all cases and wouldn't expand the option to all cases unless the performance issue is a concern.

Allowing it to be a model level property might not be helpful because the user might just want to disable default sorting on either one of `find` or `findOne` function in which case it will be disabled for both. Passing it down the function options sounds reasonable for me.

@raymondfeng Thoughts?

slnode:

slnode:

slnode:

slnode:

slnode:

rmg:
@slnode ok to test

PR #179, opened 2015-10-05
Add support for connection string options

acourtiol:


slnode:

jdrouet:
:+1:

raymondfeng:
We already have support for mongodb connection string. Just set `url` to `mongodb://...`.

acourtiol:
We can't just set the url in our case and giving users the choice wouldn't hurt. Whatever...

raymondfeng:
@acourtiol Can you elaborate your use case? Maybe an example of config?

PR #190, opened 2015-11-13
Add env variable for mongodb server

cgole:
If the MONGODB_HOST variable is set, use that instead of localhost.


rmg:
LGTM, but we should probably add the port, too, just in case it is different:
```js
port: process.env.MONGODB_PORT || 27017,
```

PR #203, opened 2015-12-09
Allow runtime configurable test environment

superkhau:
@raymondfeng Merging without review. I'm adding this to allow easy config of running tests locally/dynamically by passing in an env var instead.


superkhau:
@slnode test please

superkhau:
@raymondfeng Merging this first. Then the fix.

PR #211, opened 2015-12-19
Clean up package.json

superkhau:
- Change keywords to lowercase since lowercase is more common in npm searches
  (uppercase and lowercase searches yield different results)
- Reorder keywords alphabetically
- Reorder keys to match default scaffold (ie. npm init) ordering
- Update package description

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,20 +1,25 @@
 {
   "name": "loopback-connector-mongodb",
   "version": "1.13.2",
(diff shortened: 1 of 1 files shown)


superkhau:
@slnode test please

superkhau:
@raymondfeng Merging without review as this is just simple cleanup.

PR #212, opened 2015-12-19
Add AUTHORS file

superkhau:


superkhau:
@raymondfeng Merging without review as its just a simple addition.

PR #213, opened 2015-12-19
Update AUTHORS

superkhau:
- Fix ordering
- Add GitHub URLs


superkhau:
@raymondfeng Merging without review as this is a simple fix.

raymondfeng:
Is this a requirement? I don't want to attach work email here :-(

superkhau:
No, it's optional. Removing all emails.

PR #215, opened 2015-12-19
Remove email from AUTHORS

superkhau:


superkhau:
@raymondfeng Removing emails.

raymondfeng:
Lgtm

PR #216, opened 2015-12-24
Upgrade should to 8.0.2

superkhau:
Depends on https://github.com/strongloop/loopback-datasource-juggler/pull/804


raymondfeng:
LGTM

superkhau:
Juggler needs to be published before merging.

PR #222, opened 2016-01-11
Fix tests for windows

Amir-61:
*Fix tests for windows where `make` command is not recognized

Just testing if CI passes on other machines as well


Amir-61:
@slnode test please.

Issue #201, opened 2015-12-07
Wrong positioning of db options in lib/mongodb.js#initialize()


benkroeger:
While debugging some other issue, I found that the `initialize` function actually sets database options like `safe` and `w` (`writeConcern`) on the settings that are passed on the the native-mongo-db client's [connect](http://mongodb.github.io/node-mongodb-native/2.0/api/MongoClient.html#.connect) method.  
This method actually would expect **db options** in a hash property of the settings called `db`

``` js
settings = settings || {};
settings.db = settings.db || {};
settings.db.safe = true;
settings.db.w = 1;
```

Also, you can define other options for your MongoClient in the `datasources.json` files  

``` js
"foo": {
    "name": "foo",
    "connector": "mongodb",
    "hostname": "localhost",
    "port": 27017,
    "database": "foo-dev",
    "username": "my-user",
    "password": "shhh it's a secret!!",
    "server": {},
    "mongos": {},
    "replSet": {},
    "db": {}
  }
```

Amir-61:
@benkroeger,

Could you please clarify more if there is something wrong? As far as I can say In [initialize method](https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L56-L72) `dataSource.settings.safe` and `dataSource.settings.w` can be undefined and in this case the default values will be set.

Thanks a lot!

benkroeger:
@Amir-61 it is correct, that this module would use default values for the settings `safe` and `w`. However, the way these settings are passed on to the [MongoClient.connect](http://mongodb.github.io/node-mongodb-native/2.0/api/MongoClient.html) method is not correct. The constructor will simply ignore them, since the only valid options are `uri_decode_auth`, `db`, `server`, `replSet`, `mongos` and `promiseLibrary`.
The writeconcern settings you're intending to provide would be a sub-key in  the `db` option. I.e. you would need to provide `dataSource.settings.db.w` for the [writeconcern](https://docs.mongodb.org/v3.0/reference/write-concern/). The `safe` option on the other hand doesn't seem to be a valid option in the [mongodb-native driver](http://mongodb.github.io/node-mongodb-native/2.0/api/index.html) at all.
Might it be, that you had the [mongoose implementation](http://mongoosejs.com/docs/guide.html#safe) in mind? Mongoose uses this `safe` schema option to define [writeconcern](https://docs.mongodb.org/v3.0/reference/write-concern/) default options.

[...]

PR #233, opened 2016-02-17
add default write-concern when "safe" setting is "true

benkroeger:
connect to: #201 
The issue description for more details.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -68,8 +68,24 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
 
   var s = dataSource.settings;
 
-  s.safe = (s.safe !== false);
-  s.w = s.w || 1;
+  // create db property if it doesn't exist yet
+  s.db = s.db || {};
+
+  // define safe writeConcern default values
+  // https://docs.mongodb.org/v3.0/reference/write-concern/
+  var safeOptions = {
+    w: 1,               // write to standalone server or at least primary in replSet
+    j: true             // request acknowledgement from MongoDB that the write operation has been written to the journal
+    // wtimeout: 1000   // timeout is only applicable for w > 1
+  };
+
+  // when safe option is set to "true", assign default values for "w" and "j" to db options
+  if (s.safe === true) {
+    Object.keys(safeOptions).forEach(function (option) {
+      s.db[option] =  safeOptions[option];
+    });
+  }
+
   s.url = s.url || generateMongoDBURL(s);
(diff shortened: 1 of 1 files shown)


slnode:

benkroeger:
closes #201

jannyHou:
@benkroeger  Thank you for contribution and sorry for the delay, could you add some test cases to verify your change? We will review your pr asap after test case ready. Thanks!

slnode:

slnode:

slnode:

slnode:

jannyHou:
And after a quick review, I think this might be a breaking change, since it changes the role of `db`, it's not an alias of `database` any more but servers to contain more config info. 
Should be documented after pr landed.

lib/mongodb.js
```diff
@@ -60,8 +60,24 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
...
+  // define safe writeConcern default values
+  // https://docs.mongodb.org/v3.0/reference/write-concern/
+  var safeOptions = {
+    w: 1,               // write to standalone server or at least primary in replSet
+    j: true             // request acknowledgement from MongoDB that the write operation has been written to the journal
+    // wtimeout: 1000   // timeout is only applicable for w > 1
+  };
+
+  // when safe option is set to "true", assign default values for "w" and "j" to db options
+  if (s.safe === true) {
```
ssh24: shouldn't it be ```if (s.db.safe === true)```?

slnode:

lib/mongodb.js
benkroeger: the documented datasource settings would allow to define `dataSource.settings.safe`, which is exactly the reason for this PR. when setting `dataSource.settings.safe` to true, none of the safe options would have been forwarded / interpreted by `mongodb`. So, this PR allows to continue setting `dataSource.settings.safe` and actually get the settings activated that were intended. I assume, the original intention for this setting here was to have sth similar to what [mongoose allows with their `safe` option on the connection object](http://mongoosejs.com/docs/guide.html#safe).

so no, it shouldn't be `if (s.db.safe === true)`

benkroeger:
PR rebased on latest master

ssh24:
Labeled it as a breaking change since the config option of `db` will not work the same way anymore. This needs to go out on a major release.

@benkroeger Could you please add some test cases to verify your patch?

slnode:

PR #241, opened 2016-04-14
Lazy connect when booting app

jannyHou:
Connect to strongloop/loopback#2242
When booting app, datasources are setup with `ds.lazyConnect = true` to avoid try connection.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -67,7 +67,13 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   dataSource.ObjectID = mongodb.ObjectID;
 
   if (callback) {
-    dataSource.connector.connect(callback);
+    if (s.lazyConnect) {
+      process.nextTick(function() {
+        callback();
+      });
+    } else {
+      dataSource.connector.connect(callback);
+    }
   }
 };
 
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -4,6 +4,37 @@ var should = require('./init.js');
 
 var Superhero, User, Post, PostWithStringId, db;
 
+describe('lazyConnect', function() {
+  it('should skip connect phase (lazyConnect = true)', function(done) {
+    var ds = getDataSource({
+      host: '127.0.0.1',
+      port: 4,
+      lazyConnect: true,
+    });
+    var errTimeout = setTimeout(function() {
+      done();
+    }, 2000);
+
+    ds.on('error', function(err) {
+      clearTimeout(errTimeout);
+      done(err);
+    });
+  });
+
+  it('should report connection error (lazyConnect = false)', function(done) {
+    var ds = getDataSource({
+      host: '127.0.0.1',
+      port: 4,
+      lazyConnect: false,
+    });
+
+    ds.on('error', function(err) {
+      err.message.should.containEql('ECONNREFUSED');
+      done();
+    });
+  });
+});
+
 describe('mongodb connector', function() {
   before(function() {
     db = getDataSource();
@@ -117,7 +148,6 @@ describe('mongodb connector', function() {
       });
     });
   });
-
 
   it('should create indexes', function(done) {
     db.automigrate('User', function() {


lib/mongodb.js
```diff
@@ -66,8 +66,12 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   dataSource.connector = new MongoDB(s, dataSource);
   dataSource.ObjectID = mongodb.ObjectID;
 
-  if (callback) {
-    dataSource.connector.connect(callback);
+  if (s.lazyConnect) {
+    return true;
+  }
+
+  if ((!s.lazyConnect) && callback) {
```
raymondfeng: It's not necessary to test `s.lazyConnect` again.

lib/mongodb.js
```diff
@@ -66,8 +66,12 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   dataSource.connector = new MongoDB(s, dataSource);
   dataSource.ObjectID = mongodb.ObjectID;
 
-  if (callback) {
-    dataSource.connector.connect(callback);
+  if (s.lazyConnect) {
+    return true;
```
raymondfeng: Should we callback?

raymondfeng:
Please add a test to cover lazyConnect.

lib/mongodb.js
jannyHou: Thanks for reviewing such quickly lol. agreed, will change it.

lib/mongodb.js
jannyHou: I temporarily `return true` here since I need investigate more is it safe to execute [`postInit`](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L315-L332) without trying connection. 
If it doesn't use any results from `dataSource.connector.connect` then we can just `callback` here.
Otherwise I might need to pass in some data in `callback(err, data)`

lib/mongodb.js
```diff
@@ -67,7 +67,11 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   dataSource.ObjectID = mongodb.ObjectID;
 
   if (callback) {
-    dataSource.connector.connect(callback);
+    if (s.lazyConnect === "true") {
```
bajtos: Connectors should not need to know that the value was initially get from an ENV variable and is string. The code reading the env variable (loopback-boot) should convert the variable into the appropriate javascript type (`boolean`).

This check then becomes `if (s.lazyConnect)`.

bajtos: Connectors should not need to know that the value was initially get from an ENV variable and is string. Since this is a boolean flag, the connector should expect a boolean value, not a string. 

It's the responsibility of the code reading the env variable (e.g. in loopback-boot) to convert the variable into the appropriate javascript type (`boolean`). 

This check then becomes `if (s.lazyConnect)`.

jannyHou: Got it:)

jannyHou:
@slnode

jannyHou:
@slnode test please

jannyHou:
@bajtos Please review:) Thanks.

BTW, the second test case fails due to error msg doesn't match I think, I will see if I can only check error, without comparing the error msg.

test/mongodb.test.js
```diff
@@ -116,6 +117,31 @@ describe('mongodb connector', function() {
...
     });
+
+    it('should skip connect phase (lazyConnect = true)', function(done) {
+      var ds = getDataSource({
+        host: 'localhost',
+        port: 4,
+        lazyConnect: true
+      });
+      initialize(ds, function(err) {
+        should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

Please always use this ^^^ form, as it reports better assertion error messages.

test/mongodb.test.js
```diff
@@ -116,6 +117,31 @@ describe('mongodb connector', function() {
...
+    });
+
+    it('should report connection error (lazyConnect = false)', function(done) {
+      var ds = getDataSource({
+        host: 'localhost',
+        port: 4,
+        lazyConnect: false
+      });
+      initialize(ds, function(err) {
+        (!!err).should.be.true;
```
bajtos: I think the canonical check is `should.exist(err)`.

test/mongodb.test.js
```diff
@@ -116,6 +117,31 @@ describe('mongodb connector', function() {
...
+        err.message.should.match(/connect ECONNREFUSED 127.0.0.1:4/);
```
bajtos: This is rather fragile and may break in the future. Also not that `localhost` may get resolved into an IPv6 address.

I am proposing to relax the check to something like`err.message.should.contain('ECONNREFUSED')`

test/mongodb.test.js
```diff
@@ -116,6 +117,31 @@ describe('mongodb connector', function() {
...
+      });
+      initialize(ds, function(err) {
+        should.not.exist(err);
+        done();
+      })
+    });
+
+    it('should report connection error (lazyConnect = false)', function(done) {
+      var ds = getDataSource({
+        host: 'localhost',
```
bajtos: It's better to use `127.0.0.1` instead of localhost, to ensure we use IPv4 on all machines. Same applies to the other test above.

test/mongodb.test.js
```diff
@@ -116,8 +117,30 @@ describe('mongodb connector', function() {
...
+        lazyConnect: true,
+      });
+      initialize(ds, done);
+    });
+
+    it('should report connection error (lazyConnect = false)', function(done) {
+      var ds = getDataSource({
+        host: '127.0.0.1',
+        port: 4,
+        lazyConnect = false,
```
raymondfeng: Should be `lazyConnect: false`

test/mongodb.test.js
```diff
@@ -116,8 +117,30 @@ describe('mongodb connector', function() {
...
     });
-  });
 
+    it('should skip connect phase (lazyConnect = true)', function(done) {
+      var ds = getDataSource({
+        host: '127.0.0.1',
+        port: 4,
+        lazyConnect: true,
+      });
+      initialize(ds, done);
```
raymondfeng: I'm not sure if calling `initialize` is good idea as the constructor of DataSource already triggers it.

jannyHou:
@bajtos The connection error is an `emit error` instead of `throw error`, so I change my testing strategy to listen on the event. 
Now all tests pass:) ready for review.
Thanks

test/mongodb.test.js
```diff
@@ -1,9 +1,46 @@
 // This test written in mocha+should.js
 var semver = require('semver');
 var should = require('./init.js');
+var initialize = require('../lib/mongodb.js').initialize;
+var createDS = require('loopback-datasource-juggler').DataSource;
```
jannyHou: I will delete `initialize` and `createDS`, they are not used in code lol.

superkhau: you can remove `// This test written in mocha+should.js` too while you're at it ;)

lib/mongodb.js
```diff
@@ -67,7 +67,11 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   dataSource.ObjectID = mongodb.ObjectID;
 
   if (callback) {
-    dataSource.connector.connect(callback);
+    if (s.lazyConnect) {
+      callback(null);
```
raymondfeng: We should wrap it inside `process.nextTick()` or `setImmediate()` so that the callback happens after the function returns.

jannyHou: Thank you @raymondfeng , applied. 
hmm, is there any reason or benefit to callback after the return? Since it's already the last step in function `initialize`

bajtos: > is there any reason or benefit to callback after the return?

http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony

test/mongodb.test.js
```diff
@@ -4,6 +4,36 @@ var should = require('./init.js');
...
+describe('lazyConnect', function() {
+  it('should skip connect phase (lazyConnect = true)', function(done) {
+    var ds = getDataSource({
+      host: '127.0.0.1',
+      port: 4,
+      lazyConnect: true,
+    });
+    var errTimeout = setTimeout(function() {
+      done();
+    }, 5000);
```
bajtos: This makes the test to run 5 seconds, which is a lot. Can we make the delay smaller please?

test/mongodb.test.js
```diff
@@ -4,6 +4,36 @@ var should = require('./init.js');
...
+
+    ds.on('error', function(err) {
+      clearTimeout(errTimeout);
```
bajtos: I think you should also fail the test in this case, because the error was not supposed to be triggered, is my understanding correct?

```js
clearTimeout(errTimeout);
done(err);
```

test/mongodb.test.js
jannyHou: I remember see 5000 in one test case somewhere else (but could not remember where it is:( ).
On my local machine, the passing time is always 1000+ ms, so my suggestion is double it, which is 2000ms, to guarantee enough time to trigger the error.

test/mongodb.test.js
jannyHou: @bajtos actually the test case would fail after 5s delay, if error event fired, but `done(err)` will speed the failure of test, which makes test fail immediately when error fired.
Will add it then.

jannyHou:
@raymondfeng I applied your and miroslav's feedback, could you take a review again? Thanks:)

raymondfeng:
LGTM

Issue #242, opened 2016-04-21
Upgrade to the latest mongodb driver 2.1.16


demsey2:
I have a few issues with connection to mongodb database. After sometime my API stopped working and displays "topology was destroyed". I looked at similar issues with other nodejs frameworks and libraries and most of them point to some issues with connection. It has been fixed but "loopback-connector-mongodb" has a dependency to the old mongodb driver which also points to the old mongodb-core module.

Is there any plan to upgrade that or I can start looking at other dbs?

In case someone is wondering this is my setup
- AWS EBS Instance running loopback 2.27.0
- MongoDB 3.0 with mlab.com

Amir-61:
Please see [here](https://github.com/strongloop/loopback-connector-mongodb/pull/244#issuecomment-218712429); this is a bug in the `mongodb` module. The peer dependency is listed by `mongodb-core`, which is a dependency of `mongodb`.

Please folow the issue here: https://jira.mongodb.org/browse/NODE-612

Closing now...
Thanks!

PR #244, opened 2016-05-07
Fix warning UNMET PEER DEPENDENCY kerberos@~0.0

beeman:
As the title says, this fixes the warning you get when you `npm install` this connector.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,10 +24,9 @@
   "license": "MIT",
   "dependencies": {
     "loopback-connector": "^2.2.0",
-    "mongodb": "~2.0.35",
+    "mongodb": "^2.0.35",
     "async": "^1.0.0",
-    "debug": "^2.1.1",
-    "kerberos": "~0.0.21"
+    "debug": "^2.1.1"
   },
   "devDependencies": {
     "benchmark": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "loopback-connector": "^2.2.0",
     "mongodb": "~2.0.35",
     "async": "^1.0.0",
-    "debug": "^2.1.1"
+    "debug": "^2.1.1",
+    "kerberos": "0.0.21"
   },
   "devDependencies": {
     "benchmark": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "mongodb": "~2.0.35",
     "async": "^1.0.0",
     "debug": "^2.1.1",
-    "kerberos": "0.0.21"
+    "kerberos": "~0.0.21"
   },
   "devDependencies": {
     "benchmark": "^1.0.0",


slnode:

superkhau:
LGTM. Thanks for the contribution! ;) @bajtos For final review and merge.

bajtos:
@slnode test please

package.json
```diff
@@ -26,7 +26,8 @@
     "loopback-connector": "^2.2.0",
     "mongodb": "~2.0.35",
     "async": "^1.0.0",
-    "debug": "^2.1.1"
+    "debug": "^2.1.1",
+    "kerberos": "0.0.21"
```
bajtos: If the peer dependency says `kerberos@~0.0`, then why are we fixing `0.0.21` as our version? Can we use `"kerberos": "~0.0.21"` instead, so that we automatically update when a bugfix is released?

beeman: Makes sense @bajtos, I've updated the PR.

bajtos:
On the second look, I think this is a bug in the `mongodb` module. The peer dependency is listed by `mongodb-core`, which is a dependency of `mongodb`.

Either `mongodb-core` should move `kerberos` to regular dependencies, and if that's not possible, then `mongodb` needs to be fixed to explicitly list `kerberos` as a peer dep.

There is already an existing issue filled in mongodb issue tracker, see https://jira.mongodb.org/browse/NODE-612

slnode:

bajtos:
Quoting from JIRA:

> The dependency was changed to be a peer dependency so that the 99.99% of users who will never use kerberos would stop being confused when it tried to compile the extension and they either are missing compiler toolchains or are on windows. If you need kerberos all details about installing it is here.

Adding kerberos to the dependency chain may be a wrong thing to do.

At the same time, the npm warning is annoying and we should get rid of it. If mongodb is not willing to fix the issue, then we need to investigate what other options there are.

Let's keep this PR open to keep track of this problem.

package.json
```diff
@@ -26,7 +26,8 @@
     "loopback-connector": "^2.2.0",
     "mongodb": "~2.0.35",
```
bajtos: IIUC, the issue has been fixed in mongodb@2.1.18, see https://github.com/mongodb/node-mongodb-native/commit/51eb0aeac736b35e7397f294ec38c4c2e1e1671d. Instead of adding `kerberos` as a dependency, we should change this line to `"mongodb": "^2.1.18"`.

@raymondfeng @superkhau do you happen to know why we were using `~` operator preventing upgrades to newer minor versions? Is there any reason against switching to `^` operator?

superkhau: I have no idea why we're using ~. I have no objection to using ^ instead.

beeman: Sounds good!

Upgrading to the latest driver will fix #242

beeman:
@bajtos @superkhau I've updated the `mongodb` dependency as suggested and removed `kerberos` from the list.

bajtos:
@slnode ok to test

bajtos:
@raymondfeng from the history, I can see that you have reverted a commit similar to this patch here: https://github.com/strongloop/loopback-connector-mongodb/commit/35c537d4b36096a0076d065757c4aa09edc88388 Do you happen to remember what was the issue at that time and whether the issue is still there?

The CI build is green, which makes me believe we should be good to upgrade now. Any objections?

bajtos:
In chat, @raymondfeng said "go for it", I am going to land this pull request then.

bajtos:
Landed via 5665038, thank you @beeman for your contribution!

paulomcnally:
? Enter the data-source name: database
? Select the connector for database: MongoDB (supported by StrongLoop)
Connector-specific configuration:
? host: localhost
? port: 27017
? user: 
? password: 
? database: fijado
? Install loopback-connector-mongodb@^1.4 Yes
fijado-api@1.0.0 /home/polin/github/fijado-api
└─┬ loopback-connector-mongodb@1.15.1 
  ├── async@1.5.2 
  └─┬ mongodb@2.0.55 
    ├── es6-promise@2.1.1 
    ├── UNMET PEER DEPENDENCY kerberos@~0.0
    ├─┬ mongodb-core@1.2.31 
    │ └── bson@0.4.23 
    └─┬ readable-stream@1.0.31 
      └── isarray@0.0.1 

npm WARN mongodb-core@1.2.31 requires a peer of kerberos@~0.0 but none was installed.

Ignore this. It's solved. Thanks!

Issue #253, opened 2016-06-12
MongoError with empty updateAttributes({})


fabien:
As commented at https://github.com/strongloop/loopback-datasource-juggler/pull/772 the PR has uncovered a serious issue in how MongoDB treats 'empty' updates (turned into `$set` by `loopback-connector-mongodb`). 

Previously this (or any other error, for that matter) was silently ignored, and I still believe this should be the case with this specific error:  `MongoError: $set is empty. You must specify a field like so: {$mod: {<field>: ...}}`

It's not enough to just check if `data` is empty in `parseUpdateData`, and then omit `$set`; the current document state will not be returned as usual (you'll see undefined or default values instead of the actual existing data from the db).

/cc @raymondfeng

Amir-61:
Connect to https://github.com/strongloop/loopback-connector-mongodb/pull/254

Amir-61:
https://github.com/strongloop/loopback-connector-mongodb/pull/254 fixes the issue

PR #254, opened 2016-06-12
Fix all sorts of issues

### LINKED ISSUE #208: 11.3.1 breaks findbyId in mongo
I have a record with id "5656ff55447fe38d129a5767" 

in 11.3.0 the findById in the explorer works just fine, and returns the record
in 11.3.1 it no longer can find the record

https://strongloop/loopback-connector-mongodb@681da1f

may have introduced this problem

fabien:
I'm sorry for the mixed commit - but as a longtime Loopback user, with probably one of the most complete and deep test-suites out there - I cannot help to feel frustrated by the breaking changes and low code quality I tend to see time and again.

Here we go:
- fix `updateAttributes` (#253) so empty `$set` operations don't throw any errors (this alone caused > 70 tests to break); just ignore these updates and return early
- perform `toDatabase` early in `updateAttributes`
- `destroy` had additional (raw MongoDB) return values like `{ n: ... }` still in it; fixed it with non-clobbering assignment
- if  `enableGeoIndexing` is not set, don't attempt to create a `sphere2d` index for this; don't add index twice to `indexList`
- why did we ever stop coercing ObjectID's in `buildWhere`? Re-added to fix #208 and other `where` related issues (updates etc, not working anymore, as outlined on the comments there). 

/cc @bajtos @raymondfeng

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 node_modules
 coverage
 .idea
+!intl/
+intl/*
+!intl/en/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules
 coverage
 .idea
-intl/*
-!intl/en/
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 node_modules
 coverage
 .idea
-!intl/
 intl/*
 !intl/en/
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,33 @@
+2016-10-17, Version 1.17.0
+==========================
+
+ * Remove TEST prefix for env vars (#292) (Simon Ho)
+
+ * Add connectorCapabilities global object (Nick Duffy)
+
+ * Update translation files - round#2 (Candy)
+
+ * Update deps to loopback 3.0.0 RC (Miroslav Bajtoš)
+
+ * Remove conflict (jannyHou)
+
+ * fix maxDistance not supported in geo filter. (Vincent Wen)
+
+ * Use juggler@3 for running the tests (Miroslav Bajtoš)
+
+ * Remove !intl (jannyHou)
+
+ * Refactor (jannyHou)
+
+ * Globalization (jannyHou)
+
+ * Support patches afterwards (jannyHou)
+
+ * Use the latest compatible mongodb (jannyHou)
+
+ * Update URLs in CONTRIBUTING.md (#264) (Ryan Graham)
+
+
 2016-05-16, Version 1.15.2
 ==========================
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,5 +147,5 @@ Contributing to `loopback-connector-mongodb` is easy. In a few simple steps:
      inaccurate in any respect. Email us at callback@strongloop.com.
 ```
 
-[Google C++ Style Guide]: https://google-styleguide.googlecode.com/svn/trunk/cppguide.xml
-[Google Javascript Style Guide]: https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml
+[Google C++ Style Guide]: https://google.github.io/styleguide/cppguide.html
+[Google Javascript Style Guide]: https://google.github.io/styleguide/javascriptguide.xml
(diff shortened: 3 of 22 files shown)


bajtos:
@fabien thank you for taking care and fixing so many bugs! I am personally still not very familiar with this code base, I'll defer review to @raymondfeng, @superkhau or possibly @Amir-61.

I would really like to see unit-tests for all issues fixed, that's IMO the most reliable way how to prevent similar regression in the future.

> I cannot help to feel frustrated by the breaking changes and low code quality I tend to see time and again.

Yeah, I feel the same quite often. For me personally, it's really difficult to prevent breaking changes when there is no failing test catching the problem early.

raymondfeng:
@fabien The code LGTM. Would you like to add a test case to avoid regression in the future?

fabien:
@bajtos @raymondfeng I will see what I can do

fabien:
I've identified further implicit String/ObjectID coercions, that you'd probably want to be explicit (based on property type) instead. In order to keep backward compatibility, the new 'strict' behavior can be enabled using `strictObjectIDCoercion` (both on the connector settings, as well as per-model settings).

As a consequence, I've found that some test cases had actually skewed logic, which indeed let them pass, but wrongly so. 

This is why I changed the test here: https://github.com/strongloop/loopback-connector-mongodb/pull/254/commits/fe3495a103386ea80ac983bde2585fb5ee1c8355#diff-f44d7a6206dcaa5ddd05b31118b5c2beL1379

If you check the logic now, you'll see that it actually matches the intent (`Post` vs. `PostWithStringId`).

fabien:
@bajtos @raymondfeng @superkhau I believe that most of the fixes are now covered with a test case. That said, I wasn't able to test `enableGeoIndexing` - if someone knows how, please let me know.

BoLaMN:
+1 only thing I found was I had to manually change the transient datasource to use objectIds other wise embedded object ids were failing due to the build where statement converting all guid ids over to an object id.

fabien:
@BoLaMN could you elaborate by referencing the code or in a gist? Thanks!

BoLaMN:
I'll try put something together today, 

the where statement it would coerce any 24 digit over value to a object id even thou it was explicitly set as a string, reason I had it set to a string was strong-remoting + qs doesn't handle objectids or arrays nicely, in the end I modified strong-remoting to use 'filter=' + JSON.stringify(query.filter) instead of qs.stringify(filter) to handle it instead 

cheers

fabien:
@BoLaMN does your fork build upon this patch? How is it related? The said speed improvements sound promising though!

Can you help me pinpoint the specific issue you originally mentioned? I'm getting a bit confused ;)

bajtos:
@0candy @jannyHou @superkhau @raymondfeng Can one of you review this patch and work with @fabien to get it landed soon?

0candy:
LGTM, needs rebasing

superkhau:
@fabien Can you get this rebased? Let's land this.

jannyHou:
Hi @fabien could you rebase it? Thanks.

Amir-61:
LGTM.

@fabien Could you please rebase it on top of master? This patch is solving many issues.

fabien:
@Amir-61 I will try to look into it today

Amir-61:
@fabien 

>Is it OK if I rebase on this branch first? And then port (with someone's help here, perhaps) to 3.x?

Sure. Please rebase it on top of master and I'll take care of it... even if you are too busy, we can take care of rebasing too. Just please let us know.

Also please note your PR is against `master` branch which is for LB 3.x, we should **backport** it to `loopback-2.x` as well.

Thanks!

fabien:
@Amir-61 Rebased on master - can you handle the backport to v2 (if necessary at all)? Thanks!

Amir-61:
> @Amir-61 Rebased on master

Thee way your PR has rebased seems so weird to me... why lots of other commits from master added to this PR? @strongloop/fa-db-connectors @bajtos does it sound right to you?

fabien:
@Amir-61 yeah, I noticed this too, I thought it was just some Github thing, in how it was visualized diffed or something - I don't know how that happened! I just rebased using upstream/master 'on top of' my branch.

Let me see if I can do it differently.

fabien:
@Amir-61 could you please use #297 instead? Thanks!

Amir-61:
https://github.com/strongloop/loopback-connector-mongodb/pull/297 Landed.

Issue #256, opened 2016-06-14
Add a "disabled" property to the datasource configuration


stephenwil:
When having more than one mongo instance in use (e.g. local vs remote), it's annoying that the connector wants to connect on initialisation to all the instances defined in datasources.json.
Ideally, when swapping between the instances, I'd like to be able to add
"disabled" : "true" to the config to disable a particular datasource.

superkhau:
@stephenwil Would you like to submit a PR?

ssh24:
Closing this issue as a similar config property exists. Read the conversation here: https://github.com/strongloop/loopback-connector-mongodb/pull/257

PR #257, opened 2016-06-15
Check for connector enabled / disabled option

stephenwil:
Further to https://github.com/strongloop/loopback-connector-mongodb/issues/256, I've added a tiny check for a "enabled" option in the datasources, so that you can disable a particular datasource if you have multiple sources but want to swap between them, to allow the app to start up without failing to connect to the sources that are not in use.

 "mongodb": {
    "host": "localhost",
    "port": 27017,
    "url": "",
    "database": "users",
    "password": "",
    "name": "mongodb",
    "user": "",
    "connector": "mongodb",
    **"enabled": false**
  },

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -65,6 +65,10 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
 
   var s = dataSource.settings;
 
+  if (s.enabled === false) {
+    return;
+  }
+
   s.safe = (s.safe !== false);
   s.w = s.w || 1;
   s.url = s.url || generateMongoDBURL(s);


slnode:

stephenwil:
What happens now folks?

stephenwil:
Can someone to do the merge please?

slnode:

slnode:

stephenwil:
Hi,
 What's the next steps for this PR?

slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

stephenwil:
Sure, will do.

jannyHou:
Hi @stephenwil , any update regarding this pr?

jannyHou:
@stephenwil when you say "disable" a datasource, we already have a config variable called `lazyConnect` to defer the connection:
```json
"mongoDS": {
... ...
"lazyConnect": true
}
```

But this config has a bug: https://github.com/strongloop/loopback/issues/2850

stephenwil:
Ah, hadn't realised about the lazyConnect option - if that will stop any datasources attempting connection immediately when the application starts, for connections that will never connect given a particular configuration, then that sounds good, and we can ignore this PR.

jannyHou:
@stephenwil Hmm, actually I don't suggest use it because it will also disable all your apis working with that ds.


Ideally the goal is:
  - Only defer connection when booting app
  - APIs still work after app boot


But now APIs won't work. And it's not an easy refactor to achieve both goals...
We will use the bug I mentioned above to track the refactor.

jannyHou:
i am closing this since we already have one same config property (with bug though), thank you @stephenwil  really appreciate your contribution all the same.

Related issues: https://github.com/strongloop/loopback/issues/2850

PR #263, opened 2016-07-12
Use the latest compatible mongodb 

jannyHou:
mongodb changed their error message in version 2.2.x

This is a fix for CI failure:
```
1) lazyConnect should report connection error (lazyConnect = false):
     Uncaught AssertionError: expected 'failed to connect to server [127.0.0.1:4] on first connect' to contain 'ECONNREFUSED'
      at Assertion.fail (node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value [as containEql] (node_modules/should/lib/assertion.js:164:19)
      at DataSource.<anonymous> (test/mongodb.test.js:41:26)
      at DataSource.postInit (node_modules/loopback-datasource-juggler/lib/datasource.js:329:14)
      at lib/mongodb.js:148:19
      at node_modules/mongodb/lib/mongo_client.js:245:20
      at node_modules/mongodb/lib/db.js:227:14
      at [object Object].<anonymous> (node_modules/mongodb/lib/server.js:228:9)
      at [object Object].<anonymous> (node_modules/mongodb-core/lib/topologies/server.js:289:21)
      at [object Object].<anonymous> (node_modules/mongodb-core/lib/connection/pool.js:230:12)
      at Socket.<anonymous> (node_modules/mongodb-core/lib/connection/connection.js:150:49)
      at emitErrorNT (net.js:1265:8)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickCallback (internal/process/next_tick.js:98:9)
  ---------------------------------------------
      at Context.<anonymous> (test/mongodb.test.js:40:8)

  2) mongodb connector .ping(cb) should report connection errors:
     Uncaught AssertionError: expected 'failed to connect to server [localhost:4] on first connect' to match /connect ECONNREFUSED/
      at Assertion.fail (node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value [as match] (node_modules/should/lib/assertion.js:164:19)
      at test/mongodb.test.js:155:28
      at DataSource.<anonymous> (lib/mongodb.js:1245:7)
      at DataSource.postInit (node_modules/loopback-datasource-juggler/lib/datasource.js:329:14)
      at lib/mongodb.js:148:19
      at node_modules/mongodb/lib/mongo_client.js:245:20
      at node_modules/mongodb/lib/db.js:227:14
      at [object Object].<anonymous> (node_modules/mongodb/lib/server.js:228:9)
      at [object Object].<anonymous> (node_modules/mongodb-core/lib/topologies/server.js:289:21)
      at [object Object].<anonymous> (node_modules/mongodb-core/lib/connection/pool.js:230:12)
      at Socket.<anonymous> (node_modules/mongodb-core/lib/connection/connection.js:150:49)
      at emitErrorNT (net.js:1265:8)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickCallback (internal/process/next_tick.js:98:9)
  ---------------------------------------------
      at MongoDB.ping (lib/mongodb.js:1244:21)
      at DataSource.ping (node_modules/loopback-datasource-juggler/lib/datasource.js:2148:20)
      at Context.<anonymous> (test/mongodb.test.js:153:10)
```
http://ci.strongloop.com/job/loopback-connector-mongodb/1364/label=amazon-5/console

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "license": "MIT",
   "dependencies": {
     "loopback-connector": "^2.2.0",
-    "mongodb": "2.1.21",
+    "mongodb": "~2.1.21",
     "async": "^1.0.0",
     "debug": "^2.1.1"
   },
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "license": "MIT",
   "dependencies": {
     "loopback-connector": "^2.2.0",
-    "mongodb": "^2.0.35",
+    "mongodb": "2.1.21",
     "async": "^1.0.0",
     "debug": "^2.1.1"
   },
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -38,7 +38,7 @@ describe('lazyConnect', function() {
     });
 
     ds.on('error', function(err) {
-      err.message.should.containEql('ECONNREFUSED');
+      err.message.should.match(/connect ECONNREFUSED/);
       done();
     });
   });


Amir-61:
@slnode test please

jannyHou:
@Amir-61 @superkhau Please review the fix for CI in loopback-connector-mongodb. Thanks.

Amir-61:
LGTM!

@superkhau  getting major changes for `mongod` has advantageous and disadvantageous; the positive side is we always keep pace with the latest major changes; however if we go for minor changes (`"mongodb": "~2.0.35",` instead of `"mongodb": "^2.0.35"`) we don't run into this issue.

jannyHou:
@slnode test please

Amir-61:
Nitpick: I guess the commit message something like `Fix the expected error message` is more clear... thought?

bajtos:
Should we change `package.json` and bump up the minimum supported mongodb version to 2.2.x?

jannyHou:
@Amir-61 PTAL. Thanks :)

Amir-61:
@jannyHou Please fix the commit message to something along the changes you had, like:
```
Use mongod `V.2.1.21`

Use mongod `V.2.1.21` and change the expected error message accordingly
```
 
I'm not sure about these `nvm` failures; IIRC, someone said we should ignore them? Any thought about these failures? 

CC: @superkhau

loay:
No, I said they were fixed and passing with success in another PR once the missing connector was added (MySQL in my case). If these test cases pass on your local machine, you probably need to investigate more.

jannyHou:
@Amir-61 @loay I checked nvm failure, they all fail due to connection error so i believe the setup work on those machines for mongodb are not ready.

And mongodb@2.1.22-alpha4
```
Connection fails:  { [MongoError: getaddrinfo ENOTFOUND invalid invalid:4000]
  name: 'MongoError',
  message: 'getaddrinfo ENOTFOUND invalid invalid:4000' } 
It will be retried for the next request.
```

mongodb@2.2.0-alpha1
```
Connection fails:  { [MongoError: failed to connect to server [invalid:4000] on first connect]
  name: 'MongoError',
  message: 'failed to connect to server [invalid:4000] on first connect' } 
It will be retried for the next request.

```

So I believe the latest compatible version should be `mongodb@2.1.22-alpha4`
@Amir-61 PTAL again lol, thanks.

Amir-61:
> I checked nvm failure, they all fail due to connection error so i believe the setup work on those machines for mongodb are not ready.

Yea I knew they are not ready yet; @rmg could you confirm this as well please?

>I believe the latest compatible version should be mongodb@2.1.22-alpha4
@Amir-61 PTAL again lol, thanks.

@jannyHou I dont see you squashed the commits yet; Please squash them as one single commit with appropriate commit message something similar to the example [here](https://github.com/strongloop/loopback-connector-mongodb/pull/263#issuecomment-232471233)

jannyHou:
@Amir-61 I didn't squash it since I m not sure is the new version passing all test case, easy to regress if it doesn't. commit message is already changed to `Use the latest compatible mongodb`. If you think the patch is ok I will squash them then merge it. Thanks.

Amir-61:
Well we see they pass on other platforms except 'nvm' which needs @rmg's confirmation whether these failures are valid or not. Lets get the final LGTM from @bajtos.

superkhau:
The `nvm` thing is completely unrelated to this PR. Either way, we should confirm with @rmg for how to move forward and fix it (in another PR).

superkhau:
@jannyHou Reviewed. Please apply the feedback and ping me again for round 2.

package.json
```diff
@@ -24,7 +24,7 @@
   "license": "MIT",
   "dependencies": {
     "loopback-connector": "^2.2.0",
-    "mongodb": "^2.0.35",
+    "mongodb": "2.1.22-alpha4",
```
bajtos: Any version number string containing `-<suffix>` is a pre-release version accordig to semver, see http://semver.org/#spec-item-9:

> A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. 

I don't think it's a good idea to depend on an unstable version in a stable release of our connector.

I am proposing to use something like `~2.1.21` instead

jannyHou: Sounds good.

test/mongodb.test.js
```diff
@@ -38,7 +38,7 @@ describe('lazyConnect', function() {
     });
 
     ds.on('error', function(err) {
-      err.message.should.containEql('ECONNREFUSED');
+      err.message.should.match(/connect ECONNREFUSED/);
```
bajtos: I would personally go with `.match(/ECONNREFUSED/)`, but let's not bikeshed here too much. Pick whatever passes the tests and is not too specific.

No check will be entirely future proof, and the difference between the 70% chance and 90% chance of supporting newer mongodb versions is not worth wasting our time discussing it here.

That's my two cents, I'll refrain from more comments about this particular line of code.

jannyHou:
@superkhau I think the current error message comparison strategy is fine, especially across repos almost everywhere uses `error.msg.should.match(/error message/)`. But I am glad to discuss a better way to improve it through direct message.

@bajtos back to 2.1.21, PTAL, thanks.

package.json
```diff
@@ -24,7 +24,7 @@
   "license": "MIT",
   "dependencies": {
     "loopback-connector": "^2.2.0",
-    "mongodb": "^2.0.35",
+    "mongodb": "2.1.21",
```
bajtos: Please use `~2.1.21`. If the there are any bugfixes publised in `2.1.x` series, we want to get them automatically. Unless there is a reason for pinning down the version to 2.1.21 only?

bajtos:
I left a minor comment which may be ignored. Let's wait for @superkhau and if he doesn't respond until end of this week, then go ahead and land this fix.

test/mongodb.test.js
superkhau: Yeah, let's not bikeshed too much here. It was just my personal preference to check for the error code because that is unlikely to change. But checking for ECONNREFUSED is fine since it makes the test more clear anyways.

superkhau:
@jannyHou LGTM, the rest are nitpicks we don't need to waste time on.

superkhau:
>@superkhau getting major changes for mongod has advantages and disadvantages; the positive side is we always keep pace with the latest major changes; however if we go for minor changes ("mongodb": "~2.0.35", instead of "mongodb": "^2.0.35") we don't run into issues similar to this.

IMO, the key is to keep things stable for our users (whether or not upstream decides to break things -- sometimes by accident). I personally would prefer to use hardcoded versions (ie. 2.0.35 and update on a controlled basis), but I'm not sure how @bajtos and @raymondfeng feel about this. The con is that its manual checking, the pro is that things won't break since we don't upgrade without checking if things are still working (normally CI would do this for us but that have proven to be quite unreliable as of late).

jannyHou:
@superkhau @bajtos @Amir-61 @raymondfeng I personally prefer `~2.1.21`, since if it's broken in the future, we can quickly detect it from CI failure and change version back to a hardcode one. But if we hardcode the version now we wouldn't have a chance to get any future compatible patch, . Let's use `~2.1.21` in this pr to get CI pass first. I will keep an eye on it and regress to a fixed version if it causes compatible problem again :)

Amir-61:
>@jannyHou I dont see you squashed the commits yet

You forgot to squash your commits lol

Also please delete branch when you land the PR.

Thanks!

bajtos:
> IMO, the key is to keep things stable for our users (whether or not upstream decides to break things -- sometimes by accident). I personally would prefer to use hardcoded versions (ie. 2.0.35 and update on a controlled basis), but I'm not sure how @bajtos and @raymondfeng feel about this. The con is that it's manual checking, the pro is that things won't break since we don't upgrade without checking if things are still working (normally CI would do this for us but that have proven to be quite unreliable as of late).

I am afraid using hard-coded versions and upgrading manually whenever a new dependency version is published is unsustainable: let's say we have 30 loopback modules and each has 5 dependencies on average, that's 150 dependencies to watch & update. 

Hard-coded dependency versions also add unnecessary delay in shipping security fixes, especially when the dependency chain is deep. Imagine if there is a security problem discovered in a 2nd-level dependency of express. It will take 3 releases (2nd-level dep, 1st-levle dep, express itself) before we get to know about the problem, and then another release until a fixed version of LoopBack is available.

If users wish to have more stability, they can use npm shrinkwrap and explicitly bump up version of selected dependencies whenever they have time to do the necessary testing. No need to do that in each library/module.

superkhau:
That makes sense. Getting the end users to shrinkwrap is better than us having to ensure stability for them. :+1:

PR #275, opened 2016-08-11
Bind context where it was lost

desaroger:
I had issues ([loopback#2597](https://github.com/strongloop/loopback/issues/2597), [loopback#2603](https://github.com/strongloop/loopback/issues/2603)) where I was losing context.

One solution was the usage of cls-hooked instead of cls on loopback-context as proposed by @josieusa [here](https://github.com/strongloop/loopback-context/pull/2) and that solved the issue. But when using the mongo connector, sometimes the context was losing.

I propose to patch two functions with context if available. For me this solve all my issues. I don't know if this will have downsides, but for me works always fine and all tests are passed.

PS: This is my first PR and english isn't my first language, so sorry if there are some mistakes. :sweat_smile:

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -54,6 +54,21 @@ function generateMongoDBURL(options) {
   }
 }
 
+/*!
+ * Tries to get the context and patch the function passed as argument
+ * @param {Function} the function to patch
+ * @param {String} [scopeName=loopback] the scope name
+ * @returns {Function} the function patched
+ */
+function patchWithContext(fn, scopeName) {
+  scopeName = scopeName || 'loopback';
+  var ns = process && process.context && process.context[scopeName];
+  if (ns && ns.bind) {
+    fn = ns.bind(fn);
+  }
+  return fn;
+}
+
 /**
  * Initialize the MongoDB connector for the given data source
  * @param {DataSource} dataSource The data source instance
@@ -847,7 +862,7 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
     } else if (filter.offset) {
       cursor.skip(filter.offset);
     }
-    cursor.toArray(function(err, data) {
+    cursor.toArray(patchWithContext(function(err, data) {
       if (self.debug) {
         debug('all', model, filter, err, data);
       }
@@ -870,7 +885,7 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
       } else {
         callback(null, objs);
       }
-    });
+    }));
   }
 };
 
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

codyolsen:
@desaroger Thanks for your work on this! You've pointed me in the right direction towards a fix!

desaroger:
@codyolsen there is a plan to solve this issue in a near future and merge with master? We are using this package with the PR and I will more comfortable if it could has master installed.

My fix it's the way to solve this kind of issue or you are thinking in another form? I am fixing this kind of issue on some libraries and I am interested on learn about process.context, cls, etc.

Thanks for your response :dancers:

desaroger:
I found another point where the context was losing, when updating a model through related model. I pushed another commit to this PR.

bajtos:
I'll leave the review to @raymondfeng, @superkhau and other maintainers of this repository. 

However, I personally think this patch should be rejected. We should not be monkey-patching our code like this to support CLS. Instead, the fix should be implemented in the MongoDB client library we are using under the hood. Since cls-hooked is based on AsyncWrap that's becoming the standard part of Node.js core, I expect it's fair to ask the MongoDB client to play nicely with AsyncWrap. The same way how it already supports domains.

desaroger:
@bajtos yes I think the same, this was a desperate patch to not break our deadline.

I am a developer on Hooptap (we are using Bluemix and we are on IBM Global Entrepreneur), and our deadline is near and we needed to solve this as soon as possible, so we are using the patched version of this connector.

You know how long we would be talking until this is resolved? I am not comfortable with the idea of using my patched version instead of a official correctly-solved branch.

superkhau:
I agree with @bajtos, this is definitely going in the wrong direction. However, @raymondfeng should have the final say here.

jannyHou:
@desaroger Sorry to say that this is not a proper solution on loopback side. 
Considering your requirement, could you fork our repo and commit your workaround? 
Thanks for understanding.

slnode:

jannyHou:
I am closing it now. Please feel free to reopen it when further discussion needed. Thanks for understanding.

bajtos:
Further discussion: https://github.com/strongloop/loopback-context/issues/21

pbalan:
@desaroger I tried your patch, but it doesn't work. Can you try help me [here](https://github.com/strongloop/loopback-connector-mongodb/issues/421).

desaroger:
Hi @pbalan. It was more than a year ago when I did this patch, was a bad idea to use it then and I am sure there is better solutions out there now, please check them out.

Even so, if you insist on using this patch, probably doesn't work because some change on the library since I made this patch loses the context. I would recommend you check this PR code, basically I patch any function inside .map, .foreach etc. Try to find the line where you lost your context (on the updated version of the library) and add the patch there.

PR #276, opened 2016-08-18
Update eslint

Amir-61:
* Update eslint

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,7 @@
       "ignoreComments": true,
       "ignoreUrls": true,
       "ignorePattern": "^\\s*var\\s.+=\\s*(require\\s*\\()|(/)"
-    }]
+    }],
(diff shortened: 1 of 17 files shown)


superkhau:
LGTM

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@rmg I assume `nvm` machines are not setup yet and I can ignore their failure, is that right? Thanks!

rmg:
@Amir-61 it has mongodb running on it. Perhaps the connection information is not in the expected form, or the version is too new?

rmg:
Ah, found the problem. `cis-jenkins` is actually running the linter. The builds on ci.strongloop.com don't run `npm test` so they don't trigger the `posttest` that does all the linting.

Good news is you should be able to fix this by doing `npm test` locally.

PR #287, opened 2016-09-19
Add translated files

jannyHou:
Connect to strongloop-internal/scrum-loopback#1070

PR #288, opened 2016-09-22
Update deps to loopback 3.0.0 RC

bajtos:
Connect to strongloop-internal/scrum-loopback#1063


bajtos:
I'm assuming windows CI failures are unrelated, same approach was taken in #287.

PR #292, opened 2016-10-07
Remove TEST prefix for env vars

superkhau:
* TEST_ prefix is not used by CI anymore and there for serve no purpose

* Part of overall goal to standardize env var injection to
  MODULE_VARNAME (ie. MONGODB_HOST for example) convention

cc @bajtos @strongloop/fa-db-connectors


superkhau:
@rmg PTAL

PR #301, opened 2016-10-29
dot notation in non-embedded relations

shalkam:
added support for dot notation in non-embedded belongsTo and hasMany relations via aggregate command of MongoDB.
I used this code just to to make my project work .... not fully tested.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -6,7 +6,6 @@
 /*!
  * Module dependencies
  */
-var g = require('strong-globalize')();
 var mongodb = require('mongodb');
 var util = require('util');
 var async = require('async');
@@ -74,7 +73,7 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
 
   if (callback) {
     if (s.lazyConnect) {
-      process.nextTick(function() {
+      process.nextTick(function () {
         callback();
       });
     } else {
@@ -122,20 +121,20 @@ util.inherits(MongoDB, Connector);
  * @param {Error} err The error object
  * @param {Db} db The mongo DB object
  */
-MongoDB.prototype.connect = function(callback) {
+MongoDB.prototype.connect = function (callback) {
   var self = this;
   if (self.db) {
-    process.nextTick(function() {
+    process.nextTick(function () {
       callback && callback(null, self.db);
     });
   } else if (self.dataSource.connecting) {
-    self.dataSource.once('connected', function() {
-      process.nextTick(function() {
+    self.dataSource.once('connected', function () {
+      process.nextTick(function () {
         callback && callback(null, self.db);
       });
     });
   } else {
-    mongodb.MongoClient.connect(self.settings.url, self.settings, function(err, db) {
+    mongodb.MongoClient.connect(self.settings.url, self.settings, function (err, db) {
       if (!err) {
         if (self.debug) {
           debug('MongoDB connection is established: ' + self.settings.url);
@@ -143,7 +142,7 @@ MongoDB.prototype.connect = function(callback) {
         self.db = db;
       } else {
         if (self.debug || !callback) {
-          g.error('{{MongoDB}} connection is failed: %s %s', self.settings.url, err);
+          console.error('MongoDB connection is failed: ' + self.settings.url, err);
         }
       }
       callback && callback(err, db);
@@ -151,11 +150,11 @@ MongoDB.prototype.connect = function(callback) {
   }
 };
 
-MongoDB.prototype.getTypes = function() {
+MongoDB.prototype.getTypes = function () {
   return ['db', 'nosql', 'mongodb'];
 };
 
-MongoDB.prototype.getDefaultIdType = function() {
+MongoDB.prototype.getDefaultIdType = function () {
   return ObjectID;
 };
 
@@ -164,7 +163,7 @@ MongoDB.prototype.getDefaultIdType = function() {
  * @param {String} model Model name
  * @returns {String} collection name
  */
-MongoDB.prototype.collectionName = function(model) {
+MongoDB.prototype.collectionName = function (model) {
   var modelClass = this._models[model];
   if (modelClass.settings.mongodb) {
     model = modelClass.settings.mongodb.collection || model;
@@ -177,9 +176,9 @@ MongoDB.prototype.collectionName = function(model) {
  * @param {String} model The model name
  * @returns {*}
  */
-MongoDB.prototype.collection = function(model) {
+MongoDB.prototype.collection = function (model) {
   if (!this.db) {
-    throw new Error(g.f('{{MongoDB}} connection is not established'));
+    throw new Error('MongoDB connection is not established');
   }
   var collectionName = this.collectionName(model);
   return this.db.collection(collectionName);
@@ -191,7 +190,7 @@ MongoDB.prototype.collection = function(model) {
  * @param {String} model The model name
  * @param {Object} data The data from DB
  */
-MongoDB.prototype.fromDatabase = function(model, data) {
+MongoDB.prototype.fromDatabase = function (model, data) {
   if (!data) {
     return null;
   }
@@ -209,7 +208,7 @@ MongoDB.prototype.fromDatabase = function(model, data) {
         data[p] = data[p].toString();
       }
     } else if (data[p] && prop && prop.type && prop.type.name === 'GeoPoint' &&
-        this.settings.enableGeoIndexing === true) {
+      this.settings.enableGeoIndexing === true) {
       data[p] = {
         lat: data[p].coordinates[1],
         lng: data[p].coordinates[0],
@@ -226,7 +225,7 @@ MongoDB.prototype.fromDatabase = function(model, data) {
  * @param {String} model The model name
  * @param {Object} data The JSON data to convert
  */
-MongoDB.prototype.toDatabase = function(model, data) {
+MongoDB.prototype.toDatabase = function (model, data) {
   if (this.settings.enableGeoIndexing !== true) {
     return data;
   }
@@ -252,12 +251,13 @@ MongoDB.prototype.toDatabase = function(model, data) {
  * @param {String} command The command name
  * @param [...] params Parameters for the given command
  */
-MongoDB.prototype.execute = function(model, command) {
+MongoDB.prototype.execute = function (model, command) {
   var collection = this.collection(model);
   // Get the parameters for the given command
   var args = [].slice.call(arguments, 2);
   // The last argument must be a callback function
   var callback = args[args.length - 1];
+
   var context = {
     model: model,
     collection: collection, req: {
@@ -265,8 +265,8 @@ MongoDB.prototype.execute = function(model, command) {
       params: args,
     },
   };
-  this.notifyObserversAround('execute', context, function(context, done) {
-    args[args.length - 1] = function(err, result) {
+  this.notifyObserversAround('execute', context, function (context, done) {
+    args[args.length - 1] = function (err, result) {
       if (err) {
         debug('Error: ', err);
       } else {
@@ -280,7 +280,7 @@ MongoDB.prototype.execute = function(model, command) {
   }, callback);
 };
 
-MongoDB.prototype.coerceId = function(model, id) {
+MongoDB.prototype.coerceId = function (model, id) {
   // See https://github.com/strongloop/loopback-connector-mongodb/issues/206
   if (id == null) return id;
   var self = this;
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

fabien:
@shalkam this looks interesting. Would it be an option for you to add some basic tests?

shalkam:
@fabien I am really a newbie when it comes to testing.
Could you help pointing out where and how to create tests for this module within the loopback framework?

bajtos:
@jannyHou @Amir-61 @superkhau PTAL ☝️

jannyHou:
Hi @shalkam I understand that you need to get your stuff work asap, the quickest way is to fork our repo then do any change on the forked one. 

I still need some time to figure out your logic in this pr but in my opinion such a change should take place in https://github.com/strongloop/loopback-datasource-juggler/, probably in file `lib/include.js`.

Could you give us a simple use case to show what you are supporting and what's your expected behaviour? Thanks. I will see what I can help with.

shalkam:
Hi @jannyHou ,
The problem was discussed in details in this issue : https://github.com/strongloop/loopback/issues/517

and this is how it ended up for mongodb

https://github.com/strongloop/loopback/issues/517#issuecomment-88369880
>  strongloop/loopback-datasource-juggler#544 is adding support for dot-notation to the memory connector, to make it consistent with MongoDB.
> 
> This should hopefully fix the issue for @daankets as described in his earlier comment:
> 
> Memory in dev, mongo in test and beyond. It's the memory connector that has the issues. We use this one for our dev setup as it resets upon restart...
> It's the 'dot' notation that is working with mongo.
> Implementing this feature for non-embedded relations (e.g. hasMany, belongsTo) and supporting SQL connectors (both embedded and non-embedded relations) requires significant amount of time, which we can't invest right now.

For me I just want to use the where & order filter using a property of an included related model.

jannyHou:
@slnode test please

jannyHou:
@shalkam OK I see, thank you for the pr, can you add some test cases to verify your support for nested properties?
You can use similar test cases for connector "memory", see https://github.com/strongloop/loopback-datasource-juggler/blob/cd7bc46efbfdd7738940d09898798208731737bb/test/memory.test.js#L326-L357

shalkam:
@jannyHou So, I've been trying to build tests today and got stuck at one point.
Here is the test I am trying to build:
```javascript
  it('hasMany should support nested conditions', function(done) {
    User.create(function(e, u) {
      u.posts.create({}, function(e, p) {
        User.all({ include: 'posts' }, function(e, user) {
          console.log(user);
          done();
        });
      });
    });
  });
```
but I get this error on testing: 
```
 104 passing (3s)
  1 failing

  1) mongodb connector hasMany should support nested conditions:
     Uncaught TypeError: Cannot read property 'posts' of undefined
      at MongoDB.all (lib/mongodb.js:856:62)

```
In tests the model relations are defined this way
```javascript
    User.hasMany(Post);
    Post.belongsTo(User);
```
no json definition files of course.

but in my code I get the model relations from models json definition this way:
```javascript
var relations = this._models[model].settings.relations;
```
and it returns undefined getting the test to fail.

So, Do you have any idea how to get the model relations that are not defined in the json file? I've tried searching for an answer my but I didn't find any. Or may be I should change the way the code works, I don't know.

slnode:

jannyHou:
From the error message:
> Uncaught TypeError: Cannot read property 'posts' of undefined
      at MongoDB.all (lib/mongodb.js:856:62)

It fails due to user instance `u` not created successfully: code `User.create(function(e, u) {}` doesn't create a user. Haven't touched your relation code.

I assume that your test case is written in file [`/test/mongodb.test.js`](https://github.com/strongloop/loopback-connector-mongodb/blob/master/test/mongodb.test.js)
I pasted same code on my end in `mongodb.test.js` and it works fine.

@shalkam can you submit your testing commit, then we can discuss upon the code, easier to debug what's going wrong. Thanks.

lib/mongodb.js
```diff
@@ -816,18 +843,83 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
     query = self.buildWhere(model, filter.where);
   }
   var fields = filter.fields;
+  var relations = this._models[model].settings.relations;
```
jannyHou: @shalkam We don't contain/expose model or model instance in connector, so I believe there is no proper way to get model settings/relations here.

Instead, you can provide the config info you need in `options` when calling `User.find({your_filter}, {your_options}, cb)`, then when `find()` calls `MongoDB.prototype.all()` under the hood, you get those config info:
```js
MongoDB.prototype.all = function all(model, filter, options, callback) { 
  // get relation info you need from `options`
}
```
details see my comment in 

And I would suggest call 'User.find({your_filter}, {your_options}, cb)' in test case instead of calling `User.all()`, `all` is an alias of `find` but we usually use `find` in tests .

lib/mongodb.js
```diff
@@ -797,6 +797,33 @@ function hasNearFilter(where) {
   return false;
 }
 
+MongoDB.prototype.buildLookup = function (relation, include, model) {
+  var self = this;
```
jannyHou: Here I can see that you need 

  - relation.type
  - relation.model
  - relation.foreignKey
  - relation.through

Try provide them in `options` like:

```js
options: {
  mongodbAggregateConfig: {
    type: 'String'
    through: 'Boolean', // Seems now you don't handle the situation with a through model
    foreignKey: 'String',
    model: 'String'
  }

}
```

fabien:
Do we have an option to enable/disable this functionality on a per-connector settings basis? I believe this to be essential. I don't want this to affect existing functionality, nor unexpectedly run into security holes, for example. Any suggestions on the settings flag name?

lib/mongodb.js
shalkam: I don't know, the code is not working on using mongodb.test.js only.
When testing the api via http request I get the models relations.

lib/mongodb.js
shalkam: I think that will work we will just include the relations we wish to aggregate inside options
http://loopback.io/doc/en/lb2/Model-definition-JSON-file.html#data-source-specific-options
may be we can include mongodb aggregated relations only inside mongodb options.

shalkam: I will be testing this approach

shalkam:
@fabien The aggregate functionality will be working for selected relations only as defined in model definition file. we just didn't get to this part.

lib/mongodb.js
jannyHou: @shalkam If I understand correctly the options are applied on per query not per model right?
A model could have both `hasMany` and `belongsTo` relation:
```
User hasMany Account
Uer belongsTo Group
```
`user.accounts` uses `hasMany`
while `user.group` uses `belongsTo`

shalkam: Yes, that is correct. the current change creates a  `$lookup`  - that is left join for mongodb - for each relation that is included.
for more details refer to mongodb documentation
https://docs.mongodb.com/v3.2/reference/operator/aggregation/lookup/

jannyHou: Good. 
> http://loopback.io/doc/en/lb2/Model-definition-JSON-file.html#data-source-specific-options
may be we can include mongodb aggregated relations only inside mongodb options.

So I think we should include options in each api call instead of the model-definition-JSON-file

shalkam: @jannyHou Could you explain more  how would that be applied?

jannyHou: @shalkam Sure, let me elaborate more:
As I explained in  https://github.com/strongloop/loopback-connector-mongodb/pull/301#discussion_r89230766

When we do CRUD operations of a model we can provide config in each function call:
 ```js
User.find({where: {account.amount: 1000}}, {mongodbAggregateConfig: {// whatever config you need}}, function(err, result){});
```

Comment (shalkam) [inline lib/mongodb.js]: @jannyHou one last question. How to pass options via an api request via http to :
`http://localhost:3000/api/Posts` for example.
I've tested adding it via model's json and it's not working and tried searching but I couldn't find anything.

Comment (jannyHou) [inline lib/mongodb.js]: @shalkam Good point I overlooked the http part. Not sure if we can pass in options there.
From my understanding we can not.
@bajtos Could you help confirm here? Thanks.

And I am afraid we need to think of another strategy instead. I will put the explanation in new comment.

**Comment by jannyHou** [comment]
@shalkam When I go back to the original comment: https://github.com/strongloop/loopback/issues/517#issuecomment-88369880 I feel we should not support relation properties in this pr.
Second level property is like model `User` has a property `address`, which is an object type property like
```json
address: {
  street: "String",
  city: "String"
}
```
And you call api `User.find({where: {address.street: "a_street"}}, cb)`

lib/mongodb.js
shalkam: @jannyHou Also, let me confirm that I don't get this error on doing http request:
`Uncaught TypeError: Cannot read property 'posts' of undefined
at MongoDB.all (lib/mongodb.js:856:62)`
and if I do a:
`var relations = this._models[model].settings.relations;`
then a: 
`console.log(relations)`
I get the relations object successfully, that's only if it's called via an http request

lib/mongodb.js
jannyHou: @shalkam I am not an expert of http(sorry lol) so need some time to check why we can get it through http, but this looks like a bug to me, I might be wrong but  IMO connector should not see its attached model's definitions....

bajtos:
> When I go back to the original comment: strongloop/loopback#517 (comment) I feel we are not supposed to support relation properties.

> > Implementing this feature for non-embedded relations (e.g. hasMany, belongsTo) and supporting SQL connectors (both embedded and non-embedded relations) requires significant amount of time, which we can't invest right now.

@jannyHou We do want to support filtering on related models' properties, but so far we did not invest the necessary effort. If you can work with @shalkam to make this happen, then that would be awesome!

I am not sure how can I help you with determining the scope. Could you please write down parts that can be included or left out of the scope of this pull request, so that we can together figure out what's a "must have" and what can be excluded?

> Do we have an option to enable/disable this functionality on a per-connector settings basis? I believe this to be essential. I don't want this to affect existing functionality, nor unexpectedly run into security holes, for example. Any suggestions on the settings flag name?

@fabien Thank you for chiming in! I agree with you that it may be better to keep this feature hidden behind a feature flag (and disabled by default).

What's your take on the scope of the feature flag - should it be per-datasource (configured in `server/datasources.json`), per-model (configured in `common/models/my-model.json`) or something else?

I think the per-method-invocation flag stored in `options`, that was proposed in the discussion above, does not work, because it's the server developer who wants to make the decision, not the client making the call.

fabien:
@bajtos I think having a per-datasource setting as well as per-model makes most sense:

- per-datasource `true` + no explicit per-model setting  = `true`
- per-datasource `true` + per-model `true` = `true`
- per-datasource `true` + per-model `false` = `false`
- per-datasource `false` + per-model `true` = `false`

Finally, I agree, the whole purpose would be to be able to leverage '$lookup' to query relations (or MongoDB collections) through dot-notation. AFAIK dot-notation filtering within a single Model is already working fine.

jannyHou:
@shalkam Some update:
Sorry I was wrong, we do expose `modelDefinition` in connectors, and this is the right way to do it:
```js
 // you can replace 
 // https://github.com/shalkam/loopback-connector-mongodb/blob/4a7858ffa3a96ed7bef19cafce62f6dfb4cfc3f4/lib/mongodb.js#L846
 // with the following two lines
 var modelDef = this.getModelDefinition(model);
 var relations = modelDef.model.relations;
```
And I also changed some code in function `buildLookup()` to make sure we are getting the right property from `relation`, you might do same change accordingly for belongsTo:
```js
if (relation.type === 'hasMany'
    && typeof relation.through === 'undefined') {
    lookup = lookup.concat({
      $lookup: {
        from: relation.modelFrom.modelName,
        localField: '_id',
        foreignField: relation.modelTo.definition.idName(),
        as: include
      }
    });
  } 
``` 

Now I can get expected result in your test :) and we can continue with your original implementation.

BoLaMN:
hey guys saw this thread and thought my mongo aggregate mixin might have some code you could use for this. 

https://github.com/BoLaMN/loopback-mongo-aggregate-mixin

it works by adding 'aggregate' to the 'filter' on a /aggregate path (see example in README)

its got me out of a few tricky situations especially with embedded models  and works fine with $loopup too. 

eg 

```
where:
  "contents.id": job.id
aggregate: [
  {
    unwind: "$contents"
  }
  {
    lookup:
      from: "OrderStatus"
      localField: "contents.statusId"
      foreignField: "_id"
      as: "status"
  }
  {
    unwind: "$status"
  }
  {
    project:
      status: "$status.desc"
      id: 0
  }
]
```

cheers

sonicd300:
is this dead? any chance to implement it?

PR #303, opened 2016-10-31
Update README with correct doc links, etc

Amir-61:
* Update README with correct doc links, etc

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1142

/to: @superkhau @0candy @crandmck


crandmck:
LGTM, thanks!

Issue #251, opened 2016-06-06
random “MongoError: not authorized on XXX to execute command {YYY}” with mongo 3.2 since loopback upgrade


ebarault:
since my latest loopback upgrade I have random errors like these: 

```
2016-06-06T18:17:55.386Z pid:2458 worker:3 MongoError: not authorized on db to execute command {...someQuery...}
2016-06-06T18:17:55.386Z pid:2458 worker:3     at Function.MongoError.create (/var/lib/strong-pm/svc/1/work/a6e0bde254cbdedde6345bdf94e29e51828c5fa2.1465232211706/node_modules/loopback-connector-mongodb/node_modules/mongodb-core/lib/error.js:31:11)
2016-06-06T18:17:55.386Z pid:2458 worker:3     at queryCallback (/var/lib/strong-pm/svc/1/work/a6e0bde254cbdedde6345bdf94e29e51828c5fa2.1465232211706/node_modules/loopback-connector-mongodb/node_modules/mongodb-core/lib/cursor.js:200:36)
2016-06-06T18:17:55.386Z pid:2458 worker:3     at Callbacks.emit (/var/lib/strong-pm/svc/1/work/a6e0bde254cbdedde6345bdf94e29e51828c5fa2.1465232211706/node_modules/loopback-connector-mongodb/node_modules/mongodb-core/lib/topologies/server.js:119:3)
2016-06-06T18:17:55.386Z pid:2458 worker:3     at null.messageHandler (/var/lib/strong-pm/svc/1/work/a6e0bde254cbdedde6345bdf94e29e51828c5fa2.1465232211706/node_modules/loopback-connector-mongodb/node_modules/mongodb-core/lib/topologies/server.js:397:23)
2016-06-06T18:17:55.387Z pid:2458 worker:3     at Socket.<anonymous> (/var/lib/strong-pm/svc/1/work/a6e0bde254cbdedde6345bdf94e29e51828c5fa2.1465232211706/node_modules/loopback-connector-mongodb/node_modules/mongodb-core/lib/connection/connection.js:302:22)
2016-06-06T18:17:55.387Z pid:2458 worker:3     at emitOne (events.js:90:13)
2016-06-06T18:17:55.387Z pid:2458 worker:3     at Socket.emit (events.js:182:7)
2016-06-06T18:17:55.387Z pid:2458 worker:3     at readableAddChunk (_stream_readable.js:153:18)
[...]

PR #306, opened 2016-11-11
Update mongodb version

jannyHou:
Connect to strongloop/loopback-connector-mongodb#251

based on #251 (comment) and #251 (comment) I am updating mongodb to use latest major version 2.2.x


jannyHou:
@strongloop/fa-db-connectors PTAL.

PR #323, opened 2017-01-06
Delete extraneous `id` for replacById

Amir-61:
Connect to https://github.com/strongloop/loopback/issues/2926

This patch deletes an extraneous `id` field when `replaceById` is used. Please note this patch cannot have test cases, since all test cases should go to `loopback-datasource-juggler`

@strongloop/sq-lb-apex Since you are the owner of `loopback-datasource-juggler` and connectors, I assign this to you to review and land. Thanks!


ssh24:
@Amir-61 you can land the PR since you did it :slightly_smiling_face:

Issue #329, opened 2017-01-16
ForceId Validation error


loay:
### Bug or feature request

- [x] Bug

### Description of feature (or steps to reproduce if bug)
when using the methods replaceOrCreate or replacebyid, when the request contains and id that doesn't exist in the db, then an error should be thrown and not create the instance !

### Link to reference: 
https://github.com/strongloop/loopback-datasource-juggler/issues/1171
https://github.com/strongloop/loopback-datasource-juggler/pull/1210

PR #330, opened 2017-01-16
Fix replaceById to report an error when id not found

loay:
connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1171
connect to https://github.com/strongloop/loopback-connector-mongodb/issues/329

Reference:
this pr is a prerequisite for:  https://github.com/strongloop/loopback-datasource-juggler/pull/1210

Unit test is included in:

'loopback-datasource-juggler/test/manipulation.test.js' section  'replaceOrCreate when forceId is true' which is being called from 'test/imported.test.js'

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -1061,6 +1061,13 @@ MongoDB.prototype.replaceById = function replace(model, id, data, options, cb) {
   });
 };
 
+function errorIdNotFoundForReplace(idValue) {
+  var msg = 'Could not replace. Object with id ' + idValue + ' does not exist!';
+  var error = new Error(msg);
+  error.statusCode = error.status = 404;
+  return error;
+}
+
 /**
  * Update a model instance with id
  * @param {String} model The name of the model
@@ -1094,6 +1101,7 @@ MongoDB.prototype.replaceWithOptions = function(model, id, data, options, cb) {
       }
     } else {
       result = undefined;
+      err = errorIdNotFoundForReplace(id);
     }
     cb && cb(err, result, cbInfo);
   });


lib/mongodb.js
```diff
@@ -1055,12 +1055,26 @@ MongoDB.prototype.count = function count(model, where, options, callback) {
  */
 MongoDB.prototype.replaceById = function replace(model, id, data, options, cb) {
   if (this.debug) debug('replace', model, id, data);
+  var self = this;
   var oid = this.coerceId(model, id);
-  this.replaceWithOptions(model, oid, data, { upsert: false }, function(err, data) {
-    cb(err, data);
+  this.find(model, oid, options, function(err, foundInstance) {
```
bajtos: Same comment as in https://github.com/strongloop/loopback-connector/pull/73#pullrequestreview-16943741

You need to detect the outcome of the underlying replace command to decide whether an existing record was updated or whether there was no matching record found.

loay:
@bajtos PTAL again. Thanks.

lib/mongodb.js
```diff
@@ -1094,6 +1101,7 @@ MongoDB.prototype.replaceWithOptions = function(model, id, data, options, cb) {
       }
     } else {
       result = undefined;
+      idNotFoundForUpdate(id, cb);
```
bajtos: It's usually better to defer only the creation of the `Error` object to a helper method, and call the callback directly from here, to make it more clear what's happening under the hood.

Also there is no need to call `process.nextTick` at this point, because `this.execute` should have already placed our callback in one of the next ticks.

Last but not least, you must return early here, otherwise the callback is called twice (once by `idNotFoundForUpdate`, second time by `cb(err, result, cbInfo)` few lines below.

I am proposing the following:

```js
} else {
  result = undefined;
  err = errorIdNotFoundForReplace(id);
  // continue to `cb && cb(...)` below
}
```

bajtos:
@jannyHou you have contributed quite few commits to this connector, could you please review too?

raymondfeng:
LGTM

Issue #344, opened 2017-02-13
Package `mongodb` updated to version `2.2.23`


vshushkov:
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)

Because of updating of `mongodb` package to version `2.2.23` a new installation of `loopback-connector-mongodb` is not working... Now `mongodb` check income options very strictly https://github.com/mongodb/node-mongodb-native/blob/2.2/lib/mongo_client.js#L41 . So options like `url` or `debug` are triggers error...

A possible temporary solution is to lock `mongodb` version to `2.2.22`.

### Additional information (Node.js version, LoopBack version, etc)

Node.js -- 6.5.0
NPM -- 4.0.5

azatoth:
Filed upstream at https://jira.mongodb.org/browse/NODE-934

nachiket-p:
**For a quick solution:**
add "mongodb": "2.2.22" to your project's package.json, 
so NPM should resolve 2.2.22 instead of loopback-connector-mongodb's dependency "mongodb": "^2.2.21",

jannyHou:
Thank you all for reporting the problem and providing solution. We will take a look ASAP. The bug fails all test case and affect lots of users.

jannyHou:
Related pr in `mongodb`: https://github.com/mongodb/node-mongodb-native/commit/39102e11d6c905986b0971a49ace2afb770d87b2#diff-af7f90364fd34f588f43f92c13a92ab6R29

hyperd:
Thanks!. This saved my day

jannyHou:
We released loopback-connector-mongodb@3.0.1 and loopback-connector-mongodb@1.18.1 to fix it. Please update your connector and hope that will solve your problem.

dhmlau:
I've verified with loopback-connector-mongodb@3.0.1 and it's working.  Closing this issue.

PR #348, opened 2017-02-13
Remove invalid options

jannyHou:
Connect to https://github.com/strongloop/loopback-connector-mongodb/issues/344

In the latest version of mongodb@2.2.23, invalid options are not allowed to passed into the driver, [this pr](https://github.com/mongodb/node-mongodb-native/commit/39102e11d6c905986b0971a49ace2afb770d87b2#diff-af7f90364fd34f588f43f92c13a92ab6R29) defines all valid options.

Not sure if `mongodb` provides an api to validate an option, from the pr above that api doesn't exist, need some time to investigate a better way to do validation. As a workaround, I manually copied all valid options in to our code to do the check.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -135,7 +135,24 @@ MongoDB.prototype.connect = function(callback) {
       });
     });
   } else {
-    mongodb.MongoClient.connect(self.settings.url, self.settings, function(err, db) {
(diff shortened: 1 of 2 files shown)


jannyHou:
@raymondfeng PTAL. Thanks.

cc @loay

raymondfeng:
LGTM. Do we have a test case to ignore non-mongodb standard properties?

raymondfeng:
LGTM now with the test case.

PR #352, opened 2017-02-13
Fix readme

crandmck:
The current README does layout properly on loopback.io: https://loopback.io/doc/en/lb3/MongoDB-connector.html#properties

This PR fixes that problem and makes a few other formatting / organizational changes.


smartmouse:
LGTM

PR #353, opened 2017-02-20
Export the MongoDB function to allow subclassing

cookejames:
### Description
Export the MongoDB function to allow easy subclassing. This is [already done](https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/mysql.js#L48) in the MySQL connector and is referenced [in the docs](https://loopback.io/doc/en/lb3/Building-a-connector.html) as the best way to extend a connector.

#### Related issues
- None

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -13,6 +13,7 @@ var async = require('async');
 var Connector = require('loopback-connector').Connector;
 var debug = require('debug')('loopback:connector:mongodb');
 
+exports.ObjectID = ObjectID;
 /*!
  * Convert the id to be a BSON ObjectID if it is compatible
  * @param {*} id The id value
@@ -39,6 +40,7 @@ function ObjectID(id) {
   }
 }
 
+exports.generateMongoDBURL = generateMongoDBURL;
 /*!
  * Generate the mongodb URL from the options
  */
@@ -83,7 +85,9 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   }
 };
 
-/**
+exports.MongoDB = MongoDB;
+
+  /**
  * The constructor for MongoDB connector
  * @param {Object} settings The settings object
  * @param {DataSource} dataSource The data source instance
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -2013,6 +2013,21 @@ describe('mongodb connector', function() {
     });
   });
 
+  it('should export the MongoDB function', function() {
+    var module = require('../');
+    module.MongoDB.should.be.an.instanceOf(Function);
+  });
+
+  it('should export the ObjectID function', function() {
+    var module = require('../');
+    module.ObjectID.should.be.an.instanceOf(Function);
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

ssh24:
@cookejames Thanks for the contribution. Could you please fix the commit linter problems?
```
**************************************************
**
**  Linting commit logs
**
**  1 problems found:
**    d5b0be5 - Export the MongoDB, ObjectID and generateMongoDBURL: First line should be 50 characters or less (saw 82)
**
**************************************************
```

cookejames:
Have your linter settings changed recently? I'm not sure that my changes are causing this error. `npm run lint` is passing locally and my changes all appear to have lower character counts than this.

ssh24:
@cookejames This linter failure is for the PR itself and not the code. The first line should be less than 50 characters and the subsequent lines should be less than 72 characters. Please make sure to put an empty line after the first line before continuing on.

cookejames:
Sorted, thanks for clarifying that

ssh24:
@slnode ok to test

lib/mongodb.js
```diff
@@ -39,6 +40,7 @@ function ObjectID(id) {
   }
 }
 
+exports.generateMongoDBURL = generateMongoDBURL;
```
ssh24: Why do we need to use export on the helper functions? Correct me if I am wrong, but I think they are only used in this file. Same goes for `ObjectID`

cookejames: They are both very useful in subclassing. If you look at the `exports.initialize` function it uses generateMongoDBURL. In our case we want to alter the main MongoDB function (to allow document versioning) through subclassing however the initialise function is basically unchanged except for invoking our new subclass instead of `new MongoDB`. 

Similarly `ObjectID` is useful to have access to when overriding the `coerceId` or `buildWhere` functions.

ssh24: In that case, can we replace https://github.com/elektron-technology/loopback-connector-mongodb/blob/master/lib/mongodb.js#L74 with `dataSource.connector = MongoDB(s, dataSource);`?

cookejames: That would only work if we were injecting our new class somehow. Our line 74 looks like `  dataSource.connector = new MongoDbVersioned(s, dataSource);`. If we want to remove those two exports then the best way would probably be to simply make them properties of the MongoDB function so they are available to subclasses.

ssh24: No there could be potential problems if we take out code from the `initialize` function and put it on `MongoDB` constructor just to avoid having two other export functions. Could be a breaking change for the `initialize` function. It looks good to me how it is right now.

slnode:

slnode:

slnode:

slnode:

slnode:

PR #355, opened 2017-03-06
convert embedded relations to dot notation

BoLaMN:
### Description
Converts embedded relations data to dot notation for updating without overwriting current fields, eg. 

```
  strong-remoting:shared-method - __updateById__content - invoke with [ '58ba9dcf496ebba1fa53ab41',
  { statusId: 172 },
  [Function: callback] ] +0ms
  loopback:connector:mongodb before parsed { contents:
   [ { text: null,
       artisticDiscretion: 'exact',
       category: null,
       classification: null,
       contentType: 'default',
       packagedAt: null,
       deadlines: null,
       meta: null,
       dates: null,
       statusId: 172,
       action: null,
       webIONumber: null,
       webStartDate: null,
       webEndDate: null,
       webHeight: null,
       webWidth: null,
       landingUrl: null,
       animation: null,
       id: 58ba9dcf496ebba1fa53ab41,
       orderProductId: null,
       comments: [],
       files: [] } ] } +2ms
  loopback:connector:mongodb after parsed { '$set':
   { 'contents.0.artisticDiscretion': 'exact',
     'contents.0.contentType': 'default',
     'contents.0.statusId': 172,
     'contents.0.id': 58ba9dcf496ebba1fa53ab41 } } +1ms
```

#### Related issues

- strongloop/loopback#2797

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -477,6 +477,69 @@ MongoDB.prototype.parseUpdateData = function(model, data, options) {
   options = options || {};
   var parsedData = {};
 
+  var primitiveTypes = ['number', 'boolean', 'string'];
+
+  var isPrimitive = function(value) {
+    return primitiveTypes.indexOf(typeof value) !== -1 || util.isDate(value) || value === null || value.constructor && value.constructor.name === 'ObjectID';
+  };
+
+  var convertToDotNotation = function convertToDotNotation(value, modelClass) {
+    if (!value || isPrimitive(value)) {
+      return value;
+    }
+
+    var obj = {};
+
+    var flatten = function flatten(parent, prefix, child) {
+      if (!child) {
+        return;
+      }
+
+      if (isPrimitive(child)) {
+        if (!parent) {
+          parent = {};
+        }
+
+        parent[prefix] = child;
+
+        return;
+      }
+
+      var getPrefix = function getPrefix(key) {
+        if (!prefix) {
+          return key;
+        } else {
+          return prefix + '.' + key;
+        }
+      };
+
+      if (Array.isArray(child)) {
+        child.forEach(function (c, index) {
+          flatten(parent, getPrefix(index), c);
+        });
+      } else {
+        Object.keys(child).forEach(function (key) {
+          flatten(parent, getPrefix(key), child[key]);
+        });
+      }
+    };
+
+    var relations = modelClass.model.relations;
+
+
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@BoLaMN thank you for the pull request. I am not familiar enough with MongoDB and our connector to review this myself, ping @strongloop/sq-lb-apex.

Few general questions:

Is this change backwards compatible? When applications with existing data upgrade to a newer version of the connector, will they continue working as before?

I see that the patch changes only the `set` case, should we change loading the values too?

Are there any other side effects? For example, consider the following situation:
 - The embedded model instance has value `{color: 'red', extra: true}`
 - Users update the embedded instance to `{color: 'blue'}` (notice that `extra` is not present)
How is this handled by the current connector and how will it be handled by your fixed version?

bajtos:
@slnode ok to test

BoLaMN:
Should be backwards compatible for previous versions 

the way it works as it mentions in the related issue if you have an embedded model property settings so if data.settings only contains { colour: 'blue' } it completely overwrites the settings object destroying the 'settings.extra' property. 

the modification I made checks for the embedded relations and would convert it to $set: 'settings.colour': 'blue' so then mongo would only patch that field.

ssh24:
@BoLaMN Please add some test cases to confirm your changes in this PR. Also please take a look at the failing tests.

ssh24:
@BoLaMN Could please add some unit tests to go with this PR?

Also there is 6 failures caught by regression upon your code changes. Please take a look and fix them if this PR is to move any forward.

Link to failing tests: http://ci.strongloop.com/job/loopback-connector-mongodb/node=4.x,os=windows/2055/console

ssh24:
@BoLaMN Are there any updates on adding some test cases for this patch? There is also failing test regressed from your changes so please take a look.

BoLaMN:
it works in some cases but i had to remove it in the end as mongodb doesn't know how to cast it as an array if the path is not already generated so we end up with object arrays 

cant use $setOnInsert either as iit may be an embedsMany to the parent is already existing, i think to completely solve the issue it will require more knowledge from further up the we already pass it 

thanks for looking anyway, closing this pr off.

PR #356, opened 2017-03-08
modify execute so that lazyConnect works

phairow:
### Description

When setting lazyConnect we should be able to start the server with or without the database available. Also, if the db server goes down, requests should show the error rather than crash the server. When the db comes back up, requests should continue working again as before. Modified execute to support this.

PR #360, opened 2017-03-24
Fix lazy connect

phairow:
### Description

When setting lazyConnect we should be able to start the server with or without the database available. Also, if the db server goes down, requests should show the error rather than crash the server. When the db comes back up, requests should continue working again as before. Modified execute to support this.

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -280,31 +280,55 @@ MongoDB.prototype.toDatabase = function(model, data) {
  * @param [...] params Parameters for the given command
  */
 MongoDB.prototype.execute = function(model, command) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

phairow:
@ssh24 here is the new PR for issue #356 with all the cleanup.

ssh24:
@slnode ok to test

ssh24:
@phairow Thank you for the contribution. PR ready to land.

PR #364, opened 2017-04-06
Update node version for Travis CI

ssh24:
Remove node 0.6 and 0.8
Add node 4 and 6


kjdelisle:
Force-merging this since those old versions of Node aren't applicable or relevant.

PR #365, opened 2017-04-06
Add additional envs for node v4/v6

ssh24:


rmg:
Landing as-is since the CI failures are unrelated and this is needed for other PRs to build properly on Travis.

PR #369, opened 2017-04-14
test: use mongodb-3.2 on Travis

rmg:
The default version is 2.6, which is simply too old for some of the
features used in the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,10 @@ addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
+      - mongodb-3.2-precise
     packages:
       - g++-4.8
+      - mongodb-org-server
+      - mongodb-org-shell
 services:
   - mongodb
-before_script:
-  - mongo mydb_test --eval 'db.addUser("travis", "test");'


rmg:
Failing on Travis until strongloop/loopback-datasource-juggler#1314 makes it in to a public release.

ssh24:
@rmg Should it be the case that travis runs against the **released** juggler or the **master** juggler? IMO, it should run against the master so that it does not block a PR which uses the master. We could set the registry to the ci strongloop so it can npm install from the master juggler. Thoughts?

rmg:
I think it makes sense to leave Travis testing against the released versions.

It is a good and constant reminder that there are upstream changes that need to be released ASAP **and** a reminder that `master` is **not** for production use.

ssh24:
A new [juggler](https://github.com/strongloop/loopback-datasource-juggler/commit/432a2e15dbfd682aa5ad22ad2656120ae612888a) was released last night. 

@slnode test please

rmg:
Travis doesn't use comments to trigger rebuilds ;-)

ssh24:
@rmg Good to go? Code LGTM

rmg:
'tis a lovely shade of green :-D

Issue #363, opened 2017-03-27
RoleMapping not work well with mongoDB connector


kimkha:
I check with [this example](https://github.com/strongloop/loopback-example-access-control/blob/master/server/boot/sample-models.js)

After insert new record, I open mongoDB, and found that RoleMapping.principalId is `String` type (not `ObjectID`). The point is, when make the query, `buildWhere` function will convert that string to ObjectID. Then there is no result after that.

I think the problem is:

* The `MongoDB.create` function didn't convert String to ObjectID
* The `MongoDB.buildWhere` function always convert String to ObjectID.

So the type in MongoDB is mismatch.

Please fix.

jonilappalainen:
Noticed the same problem. Role.isInRole doesn't find the RoleMapping because of this. Need fix.

b-admike:
@kimkha thank you for your description and doing some research on the cause. I have confirmed this is a bug as well on my end.

kimkha:
I also fixed on my private repo, can I make a PR for this? I'm not sure, my code may be messy.

b-admike:
@kimkha thanks, that would be great. Please review https://github.com/strongloop/loopback-connector-mongodb/blob/master/CONTRIBUTING.md and once you have a PR, I will review and help you land it.

kimkha:
DONE, please review

b-admike:
@kimkha I have verified your fix in #370 with my version of `loopback-example-access-control`. Once you have added unit tests for your PR, we should be able to close this issue. @jonilappalainen You can also try to see if the issue is fixed for you. You can follow the following steps before running your app if the PR is not merged yet:
```
git clone git@github.com:kimkha/loopback-connector-mongodb.git kimkha-mongodb
cd kimkha-mongodb
npm link
cd /path/to/your/app
npm link loopback-connector-mongodb
```

b-admike:
@kimkha @jonilappalainen can you please verify if this issue is fixed if you set `RoleMapping.settings.strictObjectIDCoercion = true` in the `sample-models.js` boot script file while using the same version of the `loopback-connector-mongodb` you used to find the issue?

jonilappalainen:
👍  @b-admike it seems to work at least for me with that set in sample-models.js boot script. Cheers!

b-admike:
[...]

PR #370, opened 2017-04-21
Convert all ObjectIDs Convert all ObjectIDs

kimkha:
Please review.

#### Related issues

- #363 RoleMapping not work well with mongoDB connector

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -369,7 +369,7 @@ MongoDB.prototype.coerceId = function(model, id) {
  * @param {Object} data The model data
  * @returns {Object} The new model data
  */
-MongoDB.prototype.convertAllObjectID = function (model, data) {
+MongoDB.prototype.convertAllObjectID = function(model, data) {
   var self = this;
   var props = self._models[model].properties;
   for (var p in props) {
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -363,6 +363,26 @@ MongoDB.prototype.coerceId = function(model, id) {
 };
 
 /**
+ * Convert all objectIds
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model data
+ * @returns {Object} The new model data
+ */
+MongoDB.prototype.convertAllObjectID = function (model, data) {
+  var self = this;
+  var props = self._models[model].properties;
+  for (var p in props) {
+    var prop = props[p];
+    var value = data[p];
+    if (value && self.isObjectIDProperty(model, prop, value)) {
+      data[p] = ObjectID(value);
+    }
+  }
+  return data;
+};
+
+/**
  * Create a new model instance for the given data
  * @param {String} model The model name
  * @param {Object} data The model data
@@ -384,6 +404,7 @@ MongoDB.prototype.create = function(model, data, options, callback) {
     idName !== '_id' && delete data[idName];
   }
 
+  data = self.convertAllObjectID(model, data);
   data = self.toDatabase(model, data);
 
   this.execute(model, 'insert', data, { safe: true }, function(err, result) {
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

lib/mongodb.js
```diff
@@ -363,6 +363,26 @@ MongoDB.prototype.coerceId = function(model, id) {
 };
 
 /**
+ * Convert all objectIds
```
b-admike: Here you can say something like `coerce type of all object id properties to mongoDB's ObjectID`

rmg:
@slnode ok to test

b-admike:
@kimkha thank you for your contribution. code LGTM; please add some unit tests for the new function you wrote and to avoid regressions.

b-admike:
connected to #363

lib/mongodb.js
```diff
@@ -363,6 +363,26 @@ MongoDB.prototype.coerceId = function(model, id) {
...
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model data
+ * @returns {Object} The new model data
+ */
+MongoDB.prototype.convertAllObjectID = function(model, data) {
```
b-admike: Please ignore. It is the other way around and you have fixed it :-)

b-admike:
@kimkha thanks for applying the feedback. If you need help with writing a test case for this PR, let me know.

kimkha:
@b-admike I'm new with mocha, could you please write a test for me? My source code just convert String to ObjectId before saving. So, I think it'll OK if we save string and DB present as ObjectId.

bajtos:
@fabien Could you PTAL? I think this may go against the gist of your `strictObjectIDCoercion` setting.

I am also concerned that the change proposed by this pull request will break backwards compatibility. Data stored with the current version of the connector will store ObjectID-like properties differently than the data stored with the new version of the connector. As a result, people may end up with inconsistent data format in their databases.

Thoughts?

fabien:
In principle, this should be fine as long as `isObjectIDProperty` is used (which follows the `strictObjectIDCoercion` as well). As such, backwards compatibility would be maintained. But it's hard to know what's out there in the wild, of course.

fabien:
That said, wouldn't it make sense to add this logic to `toDatabase` instead of introducing a new method that is nearly always called alongside it?

b-admike:
@fabien I tested the fix in this PR against the original issue #363 and my conclusion after testing with the flag `RoleMapping.settings.strictObjectIDCoercion` in the code is that it doesn't matter what you set the flag to if you are using this patch. Thus, I agree with what @bajtos said about this PR going against the gist of the `strictObjectIDCoersion` flag and backwards compatibility. The issue #363 is fixed if we use `RoleMapping.settings.strictObjectIDCoercion = true` in the boot script for the `loopback-example-access-control` application. I would like to propose that we document what `strictObjectIDCoersion` flag does so that users are aware of it and know when to use it. @kimkha yes I can write the test, but I think we should figure out how to proceed further with this PR first.

kimkha:
OK, actually, I don't understand what all you mention... 😄 I'm new with loopback, just find an issue and I think I can help to fix it... So please discuss each other and let me know when you finish...

b-admike:
@kimkha what I am saying is that #363 works when we use `RoleMapping.settings.strictObjectIDCoercion = true` without your fix: see https://github.com/strongloop/loopback-connector-mongodb/issues/363#issuecomment-297437024. Can you give it a try?

kimkha:
@b-admike Oh, `strictObjectIDCoercion = true` works nice. So that mean we don't need to fix with this PR, right? But we should make doc clearer, I think.

b-admike:
@kimkha yeah you're right and that is my proposal as well. I have filed https://github.com/strongloop/loopback-connector-mongodb/issues/371. You can close this PR if you are satisfied and we can always revisit if necessary. Thank you for your contribution!

Issue #300, opened 2016-10-29
buildSort is not being used in .all


fabien:
We should use the `buildSort` method instead of handling the `order` inline in the `.all` method.

smartmouse:
How would you suggest the change should be?

fabien:
Yes, it's better code re-use and easier to maintain this way. Just replacing it with a call to `buildSort` will be fine.

ssh24:
Fixed in: https://github.com/strongloop/loopback-connector-mongodb/pull/376

PR #376, opened 2017-06-14
Invoke buildSort directly

ssh24:
### Description
Allow use of buildSort function directly in `MongoDB.prototype.all` rather than in line sorting.

connect to https://github.com/strongloop/loopback-connector-mongodb/issues/300

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -1038,41 +1038,10 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
     if (err) {
       return callback(err);
     }
-    var order = {};
 
     // don't apply sorting if dealing with a geo query
     if (!hasNearFilter(filter.where)) {
-      if (!filter.order) {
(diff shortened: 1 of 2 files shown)


lib/mongodb.js
```diff
@@ -1038,7 +1038,7 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
     if (err) {
       return callback(err);
     }
-    var order = {};
+    var order = self.buildSort(model, filter.order);
```
jannyHou: buildSort function and Line#1045-1050 contain same logic:
```js
if (!order) {
    var idNames = this.idNames(model);
    if (idNames && idNames.length) {
      order = idNames;
    }
  }
```
does it have to be in both place?

kjdelisle: Let's extract that into a common function if we can.

ssh24: Ah, good catch. Fixed it.

PR #380, opened 2017-06-28
Remove the hard-coded writeConcern

raymondfeng:
{w: 1} is not always valid.
See https://docs.mongodb.com/manual/reference/write-concern/#write-concern

### Description

When performing a PUT operation from Loopback app to compose for mongo instance, it throws exception:
```
{"error": {
   "name": "MongoError",
   "status": 500,
   "message": "w: 'majority' is the only valid write concern when writing to config server replica sets, got: { w: 1, wtimeout: 0 }",
   "ok": 0,
   "errmsg": "w: 'majority' is the only valid write concern when writing to config server replica sets, got: { w: 1, wtimeout: 0 }",
   "code": 2
}}
```
#### Related issues

- None

### Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+coverage
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -426,7 +426,7 @@ MongoDB.prototype.save = function(model, data, options, callback) {
 
   data = self.toDatabase(model, data);
 
-  this.execute(model, 'save', data, { w: 1 }, function(err, result) {
+  this.execute(model, 'save', data, {}, function(err, result) {
     if (!err) {
       self.setIdValue(model, data, idValue);
       idName !== '_id' && delete data._id;


raymondfeng:
@jannyHou Thanks for the review. I don't think it makes sense to split the http url in 2 lines. I'll use the admin right to merge it.

kequarles:
This pull request is associated with Bluemix support ticket https://jazzop27.rtp.raleigh.ibm.com:9443/ccm/web/projects/CloudOE#action=com.ibm.team.workitem.viewWorkItem&id=484915, which I have taken ownership of to monitor (in my current role as Gateway Director squad rep for the support interrupt role), as Raymond does not have access to jazzop27 (RTC repository).  @raymondfeng What can be communicated in the ticket (=> to the customer by support) regarding when the customer can expect to benefit from your fix?

raymondfeng:
@kequarles The PR should fix the issue described above as reported by the customer.

kequarles:
Per Slack conversation, Raymond confirmed that the fix is available in NPM, and the customer can obtain it by executing "npm i loopback-connector-mongodb@^3.2.0 --save" to update the  dependency for the loopback app.  I've updated the ticket with this info, for support to relay to the customer.

PR #381, opened 2017-06-28
Add docs on lazyConnect flag

ssh24:
### Description
Add docs on the `lazyConnect` flag for mongodb. The behaviour of the **lazyConnect** flag is as below:

- When `lazyConnect: true`
  - Mongo server is offline
    - App starts, Endpoints won't work
  - Mongo server is online
    - App starts, Endpoints will work

- When `lazyConnect: false`
  - Mongo server is offline
    - App won't start, Endpoints won't work
  - Mongo server is online
    - App starts, Endpoints will work

connect to https://github.com/strongloop-internal/scrum-apex/issues/74


jannyHou:
The `lazyConnect` flag is created for internal use actually, long story to explain, briefly, apiconnect uses it to make sure when generating swagger file, apiconnect doesn't validate the connection.

ssh24:
@jannyHou Not sure what your question is. If `lazyConnect: true`, the app will always start but defer the connection phase. If the mongodb server is then online, the endpoints will work (the user needs to hit the endpoint to establish the connection) and if the server is down the endpoints will not work (because there is no connection ¯\\_(ツ)_/¯ ).

jannyHou:
> the app will always start but defer the connection phase

I mean I am not sure does it "defer" the connection phase or just "skip" the connection phase(which means it won't do it at all), since I don't see the code that make the connection happen again... :(

ssh24:
I have a working application that satisfies this behaviour. You can try it out yourself.
- Shut down your mongodb server
- Start your app with `lazyConnect` flag set to **true** in the datasource config
- Try hitting an endpoint (it will fail)
- Start your mongodb server
- Do not kill the app, and from the existing explorer window hit an endpoint and it will work

README.md
```diff
@@ -60,6 +60,14 @@ Edit `datasources.json` to add any other additional properties that you require.
 `$pop`, `$pullAll`, `$pull`, `$pushAll`, `$push`, and `$bit`.  Default is `false`.
 - **enableGeoIndexing**: Set to `true` to enable 2dsphere indexing for model properties
 of type `GeoPoint`. This allows for indexed ```near``` queries.  Default is `false`.
+- **lazyConnect**: 
+  - Default is `false`.
+  - If set to `true`, the database instance will not be attached to the datasource and defer the connection.
+  - It will try to establish the connection automatically once users hit the endpoint.
+    - If the mongodb server is offline:
+      - The `app` will start, however, the `endpoints` will not work.
+    - If the mongodb server is online:
```
kjdelisle: Not sure this needs to be said, really. I'd just let them know that if their datasource isn't active and accepting connections at request-time, then their endpoints won't function correctly.

kjdelisle:
Does it also work in this scenario?
- Shut down mongodb server
- Start your app with `lazyConnect`
- Turn on your mongodb server _before_ making a request
- Make the first request

ssh24:
> Does it also work in this scenario?
> - Shut down mongodb server
> - Start your app with lazyConnect
> - Turn on your mongodb server before making a request
> - Make the first request

Yes.

README.md
```diff
@@ -60,6 +60,12 @@ Edit `datasources.json` to add any other additional properties that you require.
 `$pop`, `$pullAll`, `$pull`, `$pushAll`, `$push`, and `$bit`.  Default is `false`.
 - **enableGeoIndexing**: Set to `true` to enable 2dsphere indexing for model properties
 of type `GeoPoint`. This allows for indexed ```near``` queries.  Default is `false`.
+- **lazyConnect**: 
+  - Default is `false`.
+  - If set to `true`, the database instance will not be attached to the datasource and defer the connection.
+  - It will try to establish the connection automatically once users hit the endpoint.
+    - If the mongodb server is offline:
+      - The `app` will start, however, the `endpoints` will not work.
```
kjdelisle: Condense all of this into a single bullet point.
- It will try to establish the connection automatically once users hit the endpoint. If the mongodb server is offline, the `app` will start, however, the `endpoints` will not work.

README.md
```diff
@@ -60,6 +60,10 @@ Edit `datasources.json` to add any other additional properties that you require.
 `$pop`, `$pullAll`, `$pull`, `$pushAll`, `$push`, and `$bit`.  Default is `false`.
 - **enableGeoIndexing**: Set to `true` to enable 2dsphere indexing for model properties
 of type `GeoPoint`. This allows for indexed ```near``` queries.  Default is `false`.
+- **lazyConnect**: 
+  - Default is `false`.
+  - If set to `true`, the database instance will not be attached to the datasource and defer the connection.
```
b-admike: I suggest `and the connection is deferred` here

rahulkulmi:
@ssh24 

Sorry to say that i am trying both case and app will start both time when ```lazyConnect``` is true and false.

I am using 
```
"loopback": "^3.9.0",
"loopback-boot": "^2.25.0",
"loopback-connector-mongodb": "^3.2.1"
```

ssh24:
@rahulkulmi Is your Mongo server online? If that is the case having `lazyConnect` set to **true** or **false** will both result in starting the app successfully. Here is the possible matrix when enabling lazyConnect:

```
- When `lazyConnect: true`
  - Mongo server is offline
    - App starts, Endpoints won't work
  - Mongo server is online
    - App starts, Endpoints will work

- When `lazyConnect: false`
  - Mongo server is offline
    - App won't start, Endpoints won't work
  - Mongo server is online
    - App starts, Endpoints will work
```

rahulkulmi:
@ssh24 I am trying both case
1. When mongo server is online.
2. When mongo server is offline.

But one thing about mongo server, I am using mongo atlas server here i am allow and disallow server ip for testing server online and offline. 
may be ```lazyConnect``` is not working with mongo atlas server.

ssh24:
@rahulkulmi I have not heard about the **mongo atlast server** before. So I am not sure if our connector is compatible with that server. Are there any easy way to get a hold of that server instance so that I can test it out?

rahulkulmi:
@ssh24 This is the link for mongo atlas server
https://www.mongodb.com/cloud

I am not able to hold atlas server because my application on production mode.
if you find some thing else please tell me.

PR #383, opened 2017-08-01
Use stalebot on this repo

kjdelisle:
### Description

Stalebot will clean up stale issues after a specified time.
See the .github/stale.yml file for details.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,23 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - critical
+  - p1
+  - major
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
```
b-admike: How about `major` or `p1` here (if they exist)?

loay: I think Critical plays the same role as Major

.github/stale.yml
```diff
@@ -0,0 +1,21 @@
...
+  - security
+  - critical
+# Label to use when marking an issue as stale
+staleLabel: wontfix
```
virkt25: I think we should mark them as `stale` and not `wontfix`

.github/stale.yml
b-admike: We have two different labels though -_-

kjdelisle: Technically, they're two different categories of severity, critical being the highest and major the 2nd highest.

kjdelisle:
Tests already passed before and it's not under test, so merging. Thanks for your reviews.

Issue #395, opened 2017-10-26
Id of related model is not converted to an ObjectID when using the "gt" operator


luxigo:
Using comparison operators with a model object Id in a where clause works fine:
```where: {id: {gt: '58e7dc2d0d48ca5ff62362e3'}}```

However an empty result is returned when trying to to use comparison operators with a related model object id in a where clause:
```where: {someId: {gt: '58e7dc2d0d48ca5ff62362e3'}}```

luxigo:
Huh sorry it works with another model...

Any idea about how to spot the problem ?

luxigo:
Looks like it works with "belongsTo" but not with "hasOne"

bugdanov:
It appears that the id of the related model of a hasOne relation is converted to a string when using the "gt" operator with it (note the single quotes):

```loopback:connector:mongodb MongoDB: model=myModel command=find [ { someId: { '$gt': '59d8df203d162a00d46f8ace' } },```

Things are fine when using gt with the model id (its value is unquoted):
 ```loopback:connector:mongodb MongoDB: model=myModel command=find [ { _id: { '$gt': 59d8df203d162a00d46f8ace } }, ```

luxigo:
Ok it has to do with objectId coercion, you can use ObjectId() server side but when using the loopback-sdk-angular you cannot do nothing (yet) but write a remote method to work around it.

PR #398, opened 2017-11-09
Convert foreign keys to ObjectId in where filter

luxigo:
### Description

Convert foreign keys to ObjectId in where filter

#### Related issues

- connect to #395 

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -705,6 +705,7 @@ MongoDB.prototype.buildWhere = function(model, where) {
     }
 
     var prop = self.getPropertyDefinition(model, propName);
+    var propIsForeignKey = (prop === undefined);
 
     // Convert property to database column name
     k = self.getDatabaseColumnName(model, k);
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

kjdelisle:
@luxigo Please refrain from closing your PRs and opening new ones; you can make all of the changes necessary within the same pull request without needing to close it! :D

kjdelisle:
@slnode test please

lib/mongodb.js
```diff
@@ -705,6 +705,7 @@ MongoDB.prototype.buildWhere = function(model, where) {
     }
 
     var prop = self.getPropertyDefinition(model, propName);
+    var propIsForeignKey = (prop === undefined);
```
luxigo: Surely there must a better way to check if the property is really a foreign key: it could be a non-existing property name or a mistyped one.

But does StrongLoop connectors check for non-declared field names anywhere else or is it left to the appreciation of the database engine ? I bet for the latter.

luxigo: I will not update getPropertyDefinition() and isObjectIDProperty() since this patch does the job, but still it could be a cleaner solution.

luxigo:
review please

kjdelisle:
@slnode test please

lib/mongodb.js
kjdelisle: I don't think this is the way to solve this problem; the property being undefined could simply be an error.
You could use the `getModelDefinition` function to retrieve the model def and then examine the relations collection to see if the property is present in any of them, though, I admit that would be a lot more work.

kjdelisle:
@raymondfeng @bajtos What are your thoughts? I left a comment on a part of the PR, but I'd like a 2nd opinion.

bajtos:
@kjdelisle I don't know enough about juggler & mongodb to be able to provide a meaningful feedback beyond obvious question: Is this backwards compatible? Can this change break existing applications using `where` filters with conditions for foreign keys?

PR #407, opened 2018-01-11
Alias find as findById

jannyHou:
### Description

Some connectors call the function that finds instance by id as `find`only but some call it as `findById` only, I am unifying them as `findById` across all connectors.

#### Related issues

https://github.com/strongloop/loopback-datasource-juggler/pull/1459

### Checklist

### DIFF
--- a/test/connector-functions.test.js
+++ b/test/connector-functions.test.js
@@ -1,0 +1,31 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback-connector-mongodb
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// This test written in mocha+should.js
+var should = require('./init.js');
+
+describe('connector function - findById', function() {
+  var db, TestAlias, sampleId;
+  before(function(done) {
+    db = getDataSource();
+    TestAlias = db.define('TestAlias', { foo: { type: String }});
+    db.automigrate(function(err) {
+      if (err) return done(err);
+      TestAlias.create({ foo: 'foo' }, function(err, t) {
+        if (err) return done(err);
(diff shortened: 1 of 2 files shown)


rmg:
@slnode test please

jannyHou:
@slnode test please

test/connector-functions.test.js
```diff
@@ -0,0 +1,31 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback-connector-mongodb
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// This test written in mocha+should.js
+var should = require('./init.js');
```
bajtos: Please use `const`, see http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations

(According to package.json, we support only Node.js 4.x and newer, therefore `const` is allowed.)

test/connector-functions.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+describe('connector function - findById', function() {
+  var db, TestAlias, sampleId;
```
bajtos: Use `let`, not `var`.

test/connector-functions.test.js
jannyHou: I addressed the feedback but causes eslint failure, its version is too old in this repo and a upgrading causes 100+failures and could not be fixed by running a `--fix` command, so I reverted the change and opened an issue for eslint upgrade:
https://github.com/strongloop/loopback-connector-mongodb/issues/408
I am extremely willing to dig more but don't get too much time for it due to the interrupt issue and my lb-next process is too behind, will get back to it when everything's on the right track. Sorry for that.

PR #409, opened 2018-01-19
Use correct debug format

sanosom:
Valid options are not visualized when debug is active.


slnode:

kjdelisle:
@slnode test please

PR #416, opened 2018-02-15
CODEOWNERS: add nitro404

bajtos:
🎉 @nitro404 welcome on board! 🎉 

See also https://github.com/strongloop/loopback/pull/3799


bajtos:
@dhmlau 
> Actually i forgot in the other PRs, can we remove @ssh24? Or we can do it in a separate PR.

Please open a new PR for that.

Issue #403, opened 2017-12-02
$where and Javascript Server-side Code Injection


NelsonBrandao:
# Description/Steps to reproduce
MongoDB `$where` operator allows the execution of Javascript code for greater flexibility. This can be a security problem since this connector allows the `$where` operator to be defined on every `filter` param exposed by loopback and by doing so exposing the `$where` filter to the public.

This is valid for every model that extends `PersistedModel` since the usual CRUD endpoint are created with the `filter` query param for every `GET` request.

Something like this:
```
GET /api/brands?filter={"where": {"$where": "if(typeof fzzzuvz==='undefined'){var a=new Date();do{var b=new Date();}while(b-a<5000);fzzzuvz=1;}"}}
```
Allows everyone to delay their request by 5 seconds. Other versions can even cause infinite loop (more information on [here](https://media.blackhat.com/bh-us-11/Sullivan/BH_US_11_Sullivan_Server_Side_WP.pdf)

As a workaround we can disable javascript evaluation in the `$where` operator on the database it self ([here](https://medium.com/mongoaudit/how-to-disable-server-side-javascript-on-mongodb-a099fc0b8f75)) but we lose this functionality on the server as well.

A better solution would be that these operators are parsed or not exposed on the endpoints. This way they remain useful on the API side but disabled from the outside

BerkeleyTrue:
Any response from the @strongloop team on this?

NelsonBrandao:
This issue should not be closed due to inactivity since it's a very serious security bug for everyone using this framework and MongoDB

kamilml:
how can I parse this request before loopback take this code ?

in `server.js` using `app.use()` maybe?

tushar-briqsdata:
This issue should not be closed due to inactivity since it's a very serious security bug for everyone using this framework and MongoDB

dhmlau:
`pinned` it so that it won't get closed by stalebot.

virkt25:
Fix merged. Should be released shortly.

virkt25:
Published as 3.6.0

PR #452, opened 2018-08-14
fix: sanitize query by default

virkt25:
### Description

This PR adds a sanitization step to the `buildWhere` and `buildSort` function using the new `sanitizeFilter` function. 

This function accepts an option in an options object `disableSanitization` - which can be any truthy value to disable sanitization (filter passed in is returned as-is)

As per https://docs.mongodb.com/manual/core/server-side-javascript/ only `$where` and `mapReduce` properties can execute JavaScript on the Mongo Driver so `sanitizeFilter` removes those properties if present at the top level of the query object.

#### Related issues

fixes https://github.com/strongloop/loopback-connector-mongodb/issues/403

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -840,19 +840,22 @@ function idIncluded(fields, idName) {
   return true;
 }
 
-MongoDB.prototype.buildWhere = function(model, where) {
+MongoDB.prototype.buildWhere = function(model, where, options) {
   var self = this;
   var query = {};
   if (where === null || typeof where !== 'object') {
     return query;
   }
+
+  where = sanitizeFilter(where, options);
+
   var idName = self.idName(model);
   Object.keys(where).forEach(function(k) {
     var cond = where[k];
     if (k === 'and' || k === 'or' || k === 'nor') {
       if (Array.isArray(cond)) {
         cond = cond.map(function(c) {
-          return self.buildWhere(model, c);
+          return self.buildWhere(model, c, options);
         });
       }
       query['$' + k] = cond;
@@ -961,6 +964,7 @@ MongoDB.prototype.buildSort = function(model, order, options) {
     }
   }
   if (order) {
+    order = sanitizeFilter(order, options);
     var keys = order;
     if (typeof keys === 'string') {
       keys = keys.split(',');
@@ -1217,7 +1221,7 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
   var idName = self.idName(model);
   var query = {};
   if (filter.where) {
-    query = self.buildWhere(model, filter.where);
+    query = self.buildWhere(model, filter.where, options);
   }
   var fields = filter.fields;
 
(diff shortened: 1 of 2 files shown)


lib/mongodb.js
```diff
@@ -1798,6 +1803,23 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value) {
   }
 };
 
+function sanitizeFilter(filter, options) {
+  options = Object.assign({}, options);
+  if (options && options.disableSanitization) return filter;
+  if (!filter || typeof filter !== 'object') return filter;
+
+  for (const key in filter) {
+    if (key === '$where' || key === 'mapReduce') {
```
raymondfeng: `$mapReduce`?

test/mongodb.test.js
```diff
@@ -717,6 +718,50 @@ describe('mongodb connector', function() {
...
+        });
+      });
+    });
+  });
+
+  it('should not allow nested $where in where', function(done) {
+    Post.create({title: 'Post1', content: 'Post1 content'}, (err, p1) => {
+      Post.create({title: 'Post2', content: 'Post2 content'}, (err2, p2) => {
+        Post.create({title: 'Post3', content: 'Post3 data'}, (err3, p3) => {
+          Post.find({where: {conent: {where: 'function() {return this.content.contains("content")}'}}}, (err, p) => {
```
shimks: conent -> content
where -> $where

lib/mongodb.js
virkt25: According to the docs it's `mapReduce`. 

https://docs.mongodb.com/manual/core/server-side-javascript/

raymondfeng: Can the special keys nested in the filter deeper than the 1st level?

test/mongodb.test.js
```diff
@@ -717,6 +718,51 @@ describe('mongodb connector', function() {
...
+    });
+  });
+
+  it('does not execute a nested `$where`', function(done) {
+    Post.create({title: 'Post1', content: 'Post1 content'}, (err, p1) => {
+      Post.create({title: 'Post2', content: 'Post2 content'}, (err2, p2) => {
+        Post.create({title: 'Post3', content: 'Post3 data'}, (err3, p3) => {
+          Post.find({where: {content: {$where: 'function() {return this.content.contains("content")}'}}}, (err, p) => {
+            should.not.exist(err);
+            p.length.should.be.equal(0);
```
shimks: Let me know if I'm understanding this correctly: $where clause here is not executed, meaning the where filter effectively looks like this: `{where: {content: ''}}`, so `p` in this case is just an empty array.

virkt25: Well ... it tried to match `content` to the object `{$where: '...'}`

lib/mongodb.js
virkt25: Nope. As pointed out by the test case below that @shimks commented on, Mongo only executes these at the top most level and not at the property level.

raymondfeng:
@virkt25 

There are more similar cases, such as:
```json
{"where":{"$and":[{"$where":"function(){sleep(1000); return this.username.contains('test');}"}]}}
```

PR #458, opened 2018-09-07
Add support to protocol to be 'monogodb+srv'

raymondfeng:
See https://docs.mongodb.com/manual/reference/connection-string/

### Description

We don't have a way to specify `mongodb+srv://`. This PR makes it configurable.

#### Related issues

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ b benchmark benchmarks:
 
 .PHONY: l ld leak leak-detection
 l ld leak leak-detection:
+	npm i @airbnb/node-memwatch --no-save || npm i memwatch-next --no-save
 	@ITERATIONS=$(ITERATIONS) $(TESTER) leak-detection \
 		--recursive \
 		--reporter spec \
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -49,13 +49,17 @@ exports.generateMongoDBURL = generateMongoDBURL;
  * Generate the mongodb URL from the options
  */
 function generateMongoDBURL(options) {
+  // See https://docs.mongodb.com/manual/reference/connection-string/#dns-seedlist-connection-format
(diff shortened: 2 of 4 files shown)


lib/mongodb.js
```diff
@@ -49,13 +49,17 @@ exports.generateMongoDBURL = generateMongoDBURL;
  * Generate the mongodb URL from the options
  */
 function generateMongoDBURL(options) {
+  // See https://docs.mongodb.com/manual/reference/connection-string/#dns-seedlist-connection-format
+  // It can be `mongodb+srv` now.
+  var protocol = options.protocol || 'mongodb';
```
virkt25: For consistency can we make this `options.protocol = options.protocol || 'mongodb'`

package.json
```diff
@@ -26,23 +26,22 @@
...
     "strong-globalize": "^4.1.1"
   },
   "devDependencies": {
     "benchmark": "^2.1.4",
-    "bluebird": "^3.5.1",
+    "bluebird": "^3.5.2",
     "eslint": "^5.1.0",
     "eslint-config-loopback": "^10.0.0",
-    "loopback-datasource-juggler": "^3.0.0",
-    "memwatch-next": "^0.3.0",
```
virkt25: Yay :)

Makefile
```diff
@@ -19,6 +19,7 @@ b benchmark benchmarks:
 
 .PHONY: l ld leak leak-detection
 l ld leak leak-detection:
+	npm i memwatch-next --no-save
```
virkt25: Can we use `@airbnb/node-memwatch` instead here. It's compatible with Node 8/10. And then can you please update the `Leak Detection` section of the readme explaining that it only works on Node 8/10.

PR #460, opened 2018-09-07
Convert projection fields option to object

### LINKED ISSUE #459: Projection fields in find and optimizedFindOrCreate must be object

When given a fields selection object during `find` or `findOrCreate`, loopback converts it to an **array** of included fields and passes it to MongoDB library. But MongoDb library according to [find spec](http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#find) and *although deprecated* [findAndModify](http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#findAndModify) need a `projection` property that is an **object** and not an array.
However MongoDB library itself checks for provided array and converts it into an object but only in `find` methods of `collection.js` I guess since `findAndModify` is deprecated they did not bother adding this there.

In order to reproduce the problem, we need to use the  [optimizedFindOrCreate](https://github.com/strongloop/loopback-connector-mongodb/blob/ee24cd08b8ccc32711264831c71b1da628df357b/lib/mongodb.js#L1833) of Mongodb connector by setting the option `enableOptimizedFindOrCreate` to `true`  in MongoDB connector settings.
After that, issue a `findOrCreate` from a model that also includes a projection filter. The filter won't work

# Expected result

The arguments should be provided in the expected format of MongoDB driver. The current situation results in not getting any fields back, if you provide the projection document in a `findOrCreate` operation

# Additional information

linux x64 10.8.0
loopback-connector-mongodb@3.6.0

mitsos1os:
### Description
Fix provided `projection` option arguments to be provided in **object** expected format to MongoDB library. New object was created in order to avoid tampering with existing fields format that could interfere with different parts of code.

#### Related issues

- connect to #459 

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -127,6 +127,36 @@ const COMMAND_MAPPINGS = {
   estimatedDocumentCount: 'count',
 };
 
+/**
+ * Helper function to be used in {@ fieldsArrayToObj} in order for V8 to avoid re-creating a new
+ * function every time {@ fieldsArrayToObj} is called
+ *
+ * @see fieldsArrayToObj
+ * @param {object} result
+ * @param {string} field
+ * @returns {object}
+ */
+function arrayToObjectReducer(result, field) {
+  result[field] = 1;
+  return result;
+}
+
+exports.fieldsArrayToObj = fieldsArrayToObj;
+
+/**
+ * Helper function to accept an array representation of fields projection and return the mongo
+ * required object notation
+ *
+ * @param {string[]} fieldsArray
+ * @returns {Object}
+ */
+function fieldsArrayToObj(fieldsArray) {
+  if (!Array.isArray(fieldsArray)) return fieldsArray; // fail safe check in case non array object created
+  return fieldsArray.length ?
+    fieldsArray.reduce(arrayToObjectReducer, {}) :
+    {_id: 1};
+}
+
 exports.MongoDB = MongoDB;
 
 /**
@@ -1258,8 +1288,7 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
   fields = self.fromPropertyToDatabaseNames(model, fields);
 
   if (fields) {
-    // convert the array of fields into a projection object see http://mongodb.github.io/node-mongodb-native/3.0/api/Collection.html#find
-    var findOpts = {projection: fields};
+    var findOpts = {projection: fieldsArrayToObj(fields)};
     this.execute(model, 'find', query, findOpts, processResponse);
   } else {
     this.execute(model, 'find', query, processResponse);
@@ -1900,10 +1929,12 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
 
   var sort = self.buildSort(model, filter.order, options);
 
+  var projection = fieldsArrayToObj(filter.fields);
+
   this.collection(model).findOneAndUpdate(
     query,
     {$setOnInsert: data},
-    {projection: filter.fields, sort: sort, upsert: true},
+    {projection: projection, sort: sort, upsert: true},
     function(err, result) {
       if (self.debug) {
         debug('findOrCreate.callback', model, filter, err, result);
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -1406,6 +1406,68 @@ describe('mongodb connector', function() {
     });
   });
 
+  it('findOrCreate should properly support field projection (on create) - object', function() {
+    const query = {where: {title: 'Kopria post'}, fields: {comments: 1}};
+    const newData = {title: 'Kopria post', content: 'Xazo content', comments: ['comment1', 'comment2']};
+    return Post.findOrCreate(query, newData)
+      .then(function(result) {
+        const createdPost = result[0];
+        const created = result[1];
+        created.should.be.true();
+        should.not.exist(createdPost.title);
+        should.not.exist(createdPost.content);
+        createdPost.comments.should.containDeep(['comment1', 'comment2']);
+      });
+  });
+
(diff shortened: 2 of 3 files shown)


slnode:

dhmlau:
@slnode test please

raymondfeng:
@mitsos1os The `memwatch-next` issue should be fixed on master now. Can you rebase to master?

dhmlau:
@slnode test please.
Jenkins is still yellow after 5 hours.  Kicking off the test again.

dhmlau:
All tests are passed.  @raymondfeng , could you please review?  Thanks.

lib/mongodb.js
```diff
@@ -1237,8 +1256,7 @@ MongoDB.prototype.all = function all(model, filter, options, callback) {
   fields = self.fromPropertyToDatabaseNames(model, fields);
 
   if (fields) {
-    // convert the array of fields into a projection object see http://mongodb.github.io/node-mongodb-native/3.0/api/Collection.html#find
-    var findOpts = {projection: fields};
+    var findOpts = {projection: Array.isArray(fields) ? fieldsArrayToObj(fields) : fields};
```
virkt25: This can be simplified to `var findOpts = {projection: fieldsArrayToObj(fields)};` as `fireldsArrayToObj()` has a check to return as is for a non-array type.

lib/mongodb.js
```diff
@@ -1876,10 +1894,14 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
 
   var sort = self.buildSort(model, filter.order, options);
 
+  var projection = filter.fields && Array.isArray(filter.fields) ?
```
virkt25: Same simplification as above applies here.

test/mongodb.test.js
```diff
@@ -1406,6 +1406,37 @@ describe('mongodb connector', function() {
     });
   });
 
+  it('findOrCreate should properly support field projection (on create)', function() {
```
virkt25: Can you add a similar test where `fields` is passed in as an array?

test/mongodb.test.js
```diff
@@ -1406,6 +1406,37 @@ describe('mongodb connector', function() {
...
+        const createdPost = result[0];
+        const created = result[1];
+        created.should.be.true();
+        should.not.exist(createdPost.title);
+        should.not.exist(createdPost.content);
+        createdPost.comments.should.containDeep(['comment1', 'comment2']);
+      });
+  });
+
+  it('findOrCreate should properly support field projection (on find)', function() {
```
virkt25: Can you add a similar test where `fields` is passed in as an array?

lib/mongodb.js
mitsos1os: Done

lib/mongodb.js
mitsos1os: done

test/mongodb.test.js
mitsos1os: done

test/mongodb.test.js
mitsos1os: done

lib/mongodb.js
```diff
@@ -112,6 +112,25 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
...
+ * Helper function to accept an array representation of fields projection and return the mongo
+ * required object notation
+ *
+ * @param {string[]} fieldsArray
+ * @returns {Object}
+ */
+function fieldsArrayToObj(fieldsArray) {
+  if (!Array.isArray(fieldsArray)) return fieldsArray; // fail safe check in case non array object created
+  return fieldsArray.length ?
+    fieldsArray.reduce(function arrayToObjectReducer(result, field) {
```
bajtos: Can we move this function to the outer scope (outside of `fieldsArrayToObj`), so that V8 does not have to create a new function every time `fieldsArrayToObj` is called?

```js
function fieldsArrayToObj(fieldsArray) {
  // ...
}

function arrayToObjectReducer(result, field) {
  // ...
}
```

test/mongodb.test.js
```diff
@@ -1406,6 +1406,68 @@ describe('mongodb connector', function() {
     });
   });
 
+  it('findOrCreate should properly support field projection (on create) - object', function() {
+    const query = {where: {title: 'Kopria post'}, fields: {comments: 1}};
```
bajtos: I think we should run these new tests for all connectors supporting `findOrCreate` to ensure consisten behavior across databases.

I am ok to keep those tests here for short term, to get this fix landed soon.

As the next step, I would like these tests to be moved to juggler's "manipulation" test suite, see 
[test/manipulation.test.js#L1557-L1622](https://github.com/strongloop/loopback-datasource-juggler/blob/fcac276955a58c4aabcd2b602af046f45cfc73fe/test/manipulation.test.js#L1557-L1622).

mitsos1os: I agree to keep this change for the future

lib/mongodb.js
mitsos1os: You are right. Done!

lib/mongodb.js
```diff
@@ -127,6 +127,38 @@ const COMMAND_MAPPINGS = {
   estimatedDocumentCount: 'count',
 };
 
+exports.arrayToObjectReducer = arrayToObjectReducer;
```
bajtos: Let's keep `arrayToObjectReducer` as an internal implementation detail please and don't export it as public API. The less public API we have to maintain, the better for us.

dhmlau:
@slnode test please

dhmlau:
Thanks @mitsos1os  for your contribution.  Your PR has landed!

Issue #455, opened 2018-08-27
Deprecation warnings about collection APIs


b3rew:
# Description/Steps to reproduce
I am gettiing DeprecationWarning message when i try to insert data into my model
 
`(node:13177) DeprecationWarning: collection.insert is deprecated. Use insertOne, insertMany or bulkWrite instead.`

Mongo: v3.6.3
Nodejs : v8.11.2
loopback: v3.22.1
loopback-connector-mongodb: v3.6.0

PhilippeCorreges:
Same here.

ServisionFrontend:
Same here. my use "Repository" create method 
![image](https://user-images.githubusercontent.com/20966741/44832280-d8426a00-ac5c-11e8-8702-a69c0edcfb58.png)

suisun2015:
How can I fix this caution? Thanks

matteo-pennisi:
+1

janiscp:
+1

virkt25:
Would anyone be interested in submitting a PR for this? The task would involve checking to ensure the API between the deprecated / current API is the same and if not make the necessary code changes. Ensure the current test suite passes. 

See https://github.com/strongloop/loopback-connector-mongodb/pull/450/files for an example PR that changed the deprecated `count` to `countDocuments`.

Yaty:
@virkt25 I'll look into this tonight :)

Yaty:
OK for remove, see PR #461 .
insert, findAndModify, save and update seems to need more work than just updating the command :)
I'll see what I can do.

virkt25:
Fixed by #461 and released in version 3.7.0

virkt25:
Special thanks to @Yaty for the PR to update the underlying driver calls! 🙇 🎉

PR #461, opened 2018-09-07
Replace deprecated commands

Yaty:
### Description

Replace deprecated commands

#### Related issues

- connect to #455 

### Checklist

### DIFF
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -73,7 +73,7 @@ describe('lazyConnect', function() {
 
     ds.connector.execute(
       'TestLazy',
-      'insert',
+      'insertOne',
       {value: 'test value'},
       function(err, success) {
         if (err) {
@@ -100,11 +100,11 @@ describe('lazyConnect', function() {
 
     ds.connector.execute(
       'TestLazy',
-      'insert',
+      'insertOne',
       {value: 'test value'},
       function(err, success) {
         if (err) done(err);
-        var id = success.insertedIds[0];
+        var id = success.insertedId;
         ds.connector.should.have.property('db');
         ds.connector.db.should.have.property('topology');
         ds.connector.db.topology.should.have.property('isDestroyed');
@@ -805,8 +805,8 @@ describe('mongodb connector', function() {
     ) {
       Post.find(function(err, results) {
         events.should.eql([
-          'before execute insert',
-          'after execute insert',
+          'before execute insertOne',
+          'after execute insertOne',
           'before execute find',
           'after execute find',
         ]);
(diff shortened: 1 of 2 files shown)


slnode:

Yaty:
Ok I think I didn't forget anything, should be good for you to review :+1:

dhmlau:
@slnode test please

Yaty:
This test is from the juggler : https://github.com/strongloop/loopback-datasource-juggler/blob/3e5c805418416cd7b4a2a328386282d56867ba4e/test/persistence-hooks.suite.js#L1322

It is also failling locally but I was wondering if we could no modify the juggler to expect `isNewInstance` ? Otherwise I can just define `isNewInstance` to null in this PR and the problem is solved :) What do you think ?

virkt25:
Hi @Yaty, first off thank you for this amazing PR!! Much appreciated. I've reviewed the changes and everything looks good to me except we need to fix the one failing test. 

We can't modify the test in juggler as the tests are run against multiple connectors and ensure all connectors have the same behaviour. From looking at the error, we are expecting a `isNewInstance: true` but we are getting a `isNewInstance: [null]`. Given the test I think this is the relevant code that needs to be updated: https://github.com/strongloop/loopback-connector-mongodb/blob/4df3c63d0c4a0cb8d68339be7b06020cc69c1ddb/lib/mongodb.js#L569

Can you see what the `result` looks like. I suspect that the new `updateOne` operation is returning something different than what we had before and we just need to compute the value of `isNewInstance` based on the updated `result` we get back.

Yaty:
Oh yes you're right ! I wasn't thinking about other connectors :)
Fixed ! The upsert option was needed.

virkt25:
@slnode test please

virkt25:
@slnode test please

virkt25:
@slnode test please

test/mongodb.test.js
```diff
@@ -805,8 +805,8 @@ describe('mongodb connector', function() {
     ) {
       Post.find(function(err, results) {
         events.should.eql([
-          'before execute insert',
-          'after execute insert',
+          'before execute insertOne',
+          'after execute insertOne',
```
bajtos: I think this is a breaking change from the point of loopback-connector-mongodb users.

An application installing `before execute` or `after execute` hook an relying on `ctx.req.command` value `insert` will stop working after upgrade.

IMO, this change should be released as semver-major.

/cc @virkt25  @dhmlau @raymondfeng

virkt25: Had a discussion with @raymondfeng and we came to the agreement that the minor release for this is fine as `mongodb` never bumped a major version for this release. 

Should there be an issue with this impacting users in a negative way we can look into reverting this commit and releasing a new minor version -- followed by a new major version containing this fix. Haven't seen any related issues thus far. 😄

virkt25: After reading up some more on connector hooks I changed my mind, discussed this further and have opened a PR to address this.

Issue #442, opened 2018-06-27
DeprecationWarning: current URL string parser is deprecated


amenadiel:
# Description/Steps to reproduce

Running a loopback instance with `"loopback-connector-mongodb": "^3.4.4"` I get a deprecation warning:

```sh
DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect
```

Which in turn comes from a new valid option detailed in:

https://github.com/mongodb/node-mongodb-native/blob/v3.1.0/lib/mongo_client.js#L112

```js
* @param {boolean} [options.useNewUrlParser=false] Determines whether or not to use the new url parser
```
Also, this key has been added to `validOptionNames`  at

https://github.com/mongodb/node-mongodb-native/blob/v3.1.0/lib/operations/mongo_client_ops.js#L102

**However** if I manually add that option at 

https://github.com/strongloop/loopback-connector-mongodb/blob/205a3a35c3dde548f9b98af199b5edfab4d459f0/lib/mongodb.js#L145-L148

I get a ton of errors, basically because this "new url parser" doesn't infer the `replicaSet` parameter from the connection string.

So this is just a heads up to see if in a near future we'll need to introduce changes to this connector.

ingkenai:
to avoid that warning I stay in this version: "mongoose": "5.0.10",  for a while

07ASHISH:
to avoid this warning you have to pass in connect function { useNewUrlParser: true } this will remove that 
 error 
MongoClient.connect(url, { useNewUrlParser: true } );

azicchetti:
I can't connect to mlab with the 3.4.4 connector.
How can I fix that other than downgrading the package?

npn225:
ingkenai's solution works!  

For MEAN Stack users: 

1) Go to "package.json" file and change the mongoose version to "5.0.10"

2) Save this change to the "package.json" file

3) Head to the command prompt and make sure you are NOT running the app using "node" or "nodemon" commands!  

4) Enter and run the command "npm i" in order to install the packages associated to the changes you made in the "package.json" file

5) Then use "node" or "nodemon" commands to run the API without depreciation warning error!

niranjanreddy891:
Have anyone found solution for this issue?

npn225:
Don't know why you thumbed down my comment niranjanreddy891!

niranjanreddy891:
[...]

PR #417, opened 2018-02-18
Add case insensitive indexes support

angfal:
### Description

MongoDB allows to create and search by case insensitive indexes. This fixes add an additional `collation` option for find requests to support this feature.

#### Related issues

- connect to #414

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #462, opened 2018-09-08
Use new url parser by default

Yaty:
### Description

In a previous PR (#450) useNewUrlParser was added inside validOptionNames.

As it is deprecated to not use it I think by default this connector should.

(node:20938) DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.

#### Related issues

#442 

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -99,6 +99,7 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   s.safe = s.safe !== false;
   s.w = s.w || 1;
   s.url = s.url || generateMongoDBURL(s);
+  s.useNewUrlParser = s.useNewUrlParser !== false;
   dataSource.connector = new MongoDB(s, dataSource);
   dataSource.ObjectID = mongodb.ObjectID;
 


slnode:

dhmlau:
@slnode test please

Yaty:
Old parser  (3.0) : https://github.com/mongodb/node-mongodb-native/blob/3.0/lib/url_parser.js
New parser (3.0) : https://github.com/mongodb-js/mongodb-core/blob/3.0/lib/uri_parser.js

I haven't found anything telling that the new parser is backward compatible.

There is only one test : https://github.com/mongodb/node-mongodb-native/blob/a12aa15ac3eaae3ad5c4166ea1423aec4560f155/test/functional/uri_tests.js#L167

Maybe we should ask the mongodb team ?

virkt25:
Here's a great summary / answer I found stating the difference between the different parsers. The biggest thing to note is that at some point the driver will default to the `newUrlParser` ... this just gets us ahead of the curve. The only time this can break is if a port is missing in the URL. 

See. https://stackoverflow.com/a/50460871

> As for actually using the "new" connection URI options, the main restriction is to actually include the port on the connection string

bajtos:
Thank you for the explanation!

> The only time this can break is if a port is missing in the URL.

Well, this means that if an application has a datasource configured with `"url": "http://localhost"`, or perhaps a production environment configured with an environment variable like `MONGODB_URL=http://localhost`, then it will not start after our mongodb connector is upgraded to the new version.

To me, that's a breaking change we need to mitigate.

We can either release a new major version, so that LoopBack users can decide when they are ready for upgrade; or alternatively find a way how to fill the default port when it was not provided by the user and thus bridge the gap between the old and the new parser.

Personally, I prefer to publish a new semver-major version. Parsing MongoDB connection strings can be tricky, there may be other edge cases that differ between the old and the new parser implementation.

virkt25:
> Well, this means that if an application has a datasource configured with "url": "http://localhost", or perhaps a production environment configured with an environment variable like MONGODB_URL=http://localhost, then it will not start after our mongodb connector is upgraded to the new version.

I can't imagine someone running Mongo on the default PORT tbh ... especially in a production environment. That said I do understand your concern. 

However my concern is that the current version of the connector will eventually break for users when mongodb decides to remove the deprecation warning and switch to the newer parser as the only option anyways. Unless they do it in a semver-major version, in which case I'm not sure why they would add deprecation warning to the current version ... our users will be impacted by this sooner or later.

raymondfeng:
Do we even allow `http://localhost/mydb` for mongodb url? I think we use `mongodb://...`.

virkt25:
> Do we even allow http://localhost/mydb for mongodb url? I think we use mongodb://....

Isn't the value user supplied? If we construct it then we do so with `mongodb://` but if a user sets `url` it supersedes us constructing a URL.

bajtos:
> However my concern is that the current version of the connector will eventually break for users when mongodb decides to remove the deprecation warning and switch to the newer parser as the only option anyways. Unless they do it in a semver-major version, in which case I'm not sure why they would add deprecation warning to the current version ... our users will be impacted by this sooner or later.

Here is my understanding of the process of making a big breaking change:
 1. Implement the new behavior behind a feature flag (or as a new API). This is a semver-minor change.
 2. After some time passes, add a deprecation warning to nudge people to switch to the new behavior. This is a semver-minor change.
 3. After some more time passes, enable the new behavior by default. This is a semver-major change.
 4. Remove the old behavior. This is a semver-major change that can be done together with 3.

We have used this process in LoopBack ourselves, Express did it when removing built-in middleware too. From what I've been told, React is doing the same, although I am not sure if they print deprecation warnings or mention the deprecation in docs & release notes only.

I would be surprised if the mongodb connector was not following the same process.

Having said that, if we think it's highly unlikely to have users depending on the default port (or any other behavior provided by the old parser), and there is a flag that allows them to restore the old behavior (i.e. `useNewUrlParser: false`), then I am ok to follow the majority and have this released as semver-minor.

> Do we even allow `http://localhost/mydb` for mongodb url? I think we use `mongodb://....`

Sorry for the confusion, I forgot that MongoDB connection strings use `mongodb` schema instead of `http`. In my comment, I meant `mongodb://localhost/mydb`.

For reference, here is the specification of the connection string format:
 - https://docs.mongodb.com/manual/reference/connection-string/
 - https://mongodb.github.io/node-mongodb-native/driver-articles/mongoclient.html

Interestingly enough, the docs says that `port` is optional. It makes me wonder why the new URL parser requires the port to be present. Is it perhaps an oversight that can/will be fixed in a later version of the client package? Can we contribute the fix ourselves?

virkt25:
> Interestingly enough, the docs says that port is optional.

I think this might be the case since the `newUrlParser` which requires the port to be present is an `opt-in` for now. They will likely update the docs once it becomes the default. 

Considering the discussion I agree that for us to enable the `newUrlParser` by default is a breaking change and would require a major version -- we can do that now or wait for Mongo to release their new major version making this the default at which point we can release a new major version with the dependency bump on our end and no code changes.

bajtos:
> we can do that now or wait for Mongo to release their new major version making this the default at which point we can release a new major version with the dependency bump on our end and no code changes.

I am fine either way, with a slight preference for releasing a new major version now, to get rid of the deprecation warning printed by the MongoDB client.

Yaty:
So what's your decision about this PR ? We're still annoyed by the warning :smile:

virkt25:
Team will be releasing a new major version shortly — gathering a few PRs that require a major version bump together such as https://github.com/strongloop/loopback-connector-mongodb/pull/417 and dropping support for older version of Mongo.

dhmlau:
@slnode test please

jannyHou:
I upgraded the mongodb server from 3.2 to 3.4 on jenkins, rebased this PR to trigger a new build.

dhmlau:
@Yaty , your changes in this PR is in the newly released `loopback-connector-mongodb@4.0.0`.  Thanks for your contribution.

PR #465, opened 2018-09-14
fix: map new names to old for connector hooks

virkt25:
### Description

Discussed https://github.com/strongloop/loopback-connector-mongodb/pull/461#r217335767 further with @raymondfeng who suggested mapping the new names we emit in the observer to the old ones to prevent this from being a breaking change.

#### Related issues

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -112,6 +112,20 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   }
 };
 
+// MongoDB has deprecated some commands. To preserve
+// compatibility of model connector hooks, this maps the new
+// commands to previous names for the observors of this command.
+const COMMAND_MAPPINGS = {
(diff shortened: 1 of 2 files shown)


lib/mongodb.js
```diff
@@ -436,6 +436,32 @@ MongoDB.prototype.execute = function(model, command) {
       return;
     }
 
+    // MongoDB has deprecated some commands. To preserve
+    // compatibility of model connector hooks, this maps the new
+    // commands to previous names for the observors of this command.
+    switch (context.req.command) {
```
bajtos: Can we use a look-up table instead of a switch statement?

```js
// somewhere above this method

const COMMAND_MAPPING = {
  insertOne: 'insert',
  // ...
};

// here:
const mapped = COMMAND_MAPPING[context.req.command];
if (mapped) context.req.command = mapped;
```

Issue #479, opened 2018-10-23
Mongodb connector 3.9.0 hangs on various calls for loopback persisted Models


tflowcamdev:
# Description/Steps to reproduce

The 3.9.0 version suddenly hangs for us when doing things like Model.create() or Model.upsertWithWhere().

Very simple example using a persisted model called "DataType":

```
console.log('  Calling DataType.create() ...');
  DataType.create(requestData, function(error, dtInstance) {
     // Never gets here 
     console.log('  Calling DataType.create() DONE');
     return callback(error, dtInstance);
   });
```

# Additional information

The issue appears to be related to using util.inspect with depth: null in debug statements
(e.g. here:  https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L2005)

Strongloop creates objects that have nested properties - for example a persisted model will have sub properties that link to all the models (including itself).  So traversing the DataType model properties to infinite depth results in a loop.

I don't believe we're doing anything special to produce this scenario, think this is common to strongloop's persisted models.  Using utils.inspect with a smaller depth should fix the issue.

jannyHou:
@tflowcamdev Good catch, do you run into the same error with other models? Or only the model named `DataType`?

epostea:
@jannyHou We hit this issue with other models, the issue is not specific to the `DataType` model

tflowcamdev:
Yes, confirmed - This is not specific to any of our models as far as I can tell.

nursultan156:
meet the same problem.
i call:
```
CalculationM.create(calculation, {userId: settings.userId}, (e2, r2) => {
	// this never happen
          if (e2) return e(resultFactory.internal_error(e2));
          return r(resultFactory.success(null, r2));
        });
```

memory leak occurs:
```
<--- Last few GCs --->

[1:0x41460e0]   108243 ms: Mark-sweep 1410.1 (1462.9) -> 1410.1 (1446.9) MB, 3184.4 / 0.0 ms  (+ 0.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 3184 ms) last resort GC in old space requested
[1:0x41460e0]   111318 ms: Mark-sweep 1410.1 (1446.9) -> 1410.1 (1446.9) MB, 3074.6 / 0.0 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x1ad9c59a5ec1 <JSObject>
[...]

PR #480, opened 2018-10-24
remove the infinite inspect

jannyHou:
### Description

Fix https://github.com/strongloop/loopback-connector-mongodb/issues/479
The default `depth` is 4, not infinite.
#### Related issues

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -431,7 +431,7 @@ MongoDB.prototype.toDatabase = function(model, data) {
   convertDecimalProps(data, props);
   // Override custom column names
   data = this.fromPropertyToDatabaseNames(model, data);
-  debug('toDatabase data: ', util.inspect(data, {depth: null}));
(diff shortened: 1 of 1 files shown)


lib/mongodb.js
```diff
@@ -431,7 +431,7 @@ MongoDB.prototype.toDatabase = function(model, data) {
   convertDecimalProps(data, props);
   // Override custom column names
   data = this.fromPropertyToDatabaseNames(model, data);
-  debug('toDatabase data: ', util.inspect(data, {depth: null}));
+  debug('toDatabase data: ', util.inspect(data));
```
raymondfeng: This is the case that we should use: 
```ts
if (debug.enabled) debug('toDatabase data: ', util.inspect(data));
```

Otherwise, `util.inspect` is always executed, which is pretty expensive even when debug is not turned on.

Issue #457, opened 2018-09-03
How to create MongoDB entry with "NumberDecimal" Format i.e Decimal128 Data Type


gobisankaran:
# Description/Steps to reproduce
Clarification on NumberDecimal creation with MongoDB using Strongloop

I have created a simple Model in which I want to Insert a NumberDecimal datatype
When I try to create a model with Decimal128 it says "Swagger: skipping unknown type "Decimal128"."

Can you please help me to insert a sample value which should create a entry in MongoDB database with datatype "NumberDecimal i.e Decimal128" and retrieve in the same format?

Is it supported right now with loopback connector?

My sample model file looks like below
{
  "name": "DTM",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "NumDec": {
      "type": "Decimal128"
    }
  },
  "validations": [],
  "relations": {},
  "acls": [],
  "methods": {}
}

Thanks in Advance.

Regards,
Gobi S.

# Link to reproduction sandbox

# Expected result

# Additional information

virkt25:
The type should just be `number` as MongoDB doesn't have the concept of a `DECIMAL128` AFAIK.

gobisankaran:
Basically My sample model looks like this
{
  "name": "Decimal",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "NumDec": {
      "type": "object",
      "mongoIIB": {
        "dataType": "Decimal128"
      }
    }
  },
  "validations": [],
  "relations": {},
  "acls": [],
  "methods": {}
}

I want to insert a data into MongoDB as ""NumberDecimal" Data type.

If I use the above model it is inserting the value as either Integer or Double. I need to insert object as "NumberDecimal" i.e Decimal128 DataType.

Please refer this link for NumberDecimal Datatype : https://docs.mongodb.com/manual/core/shell-types/

virkt25:
LoopBack doesn't have a `decimal128` type. Here's a list of supported types in LoopBack: https://loopback.io/doc/en/lb3/LoopBack-types.html 

This connector at this time does not have support for `dataType: decimal` which is present in `sql` connectors.

raymondfeng:
A few steps need to happen
1. Check if mongodb driver accepts and coerces strings to decimal 128
[...]

PR #483, opened 2018-10-25
support nested property

jannyHou:
### Description

connect to https://github.com/strongloop/loopback-connector-mongodb/issues/457
Still in progress.
Messy code, not ready for review.

#### Related issues

### Checklist


raymondfeng:
@jannyHou Great work! Please make sure CI is good.

lib/mongodb.js
```diff
@@ -1996,17 +1991,50 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+    if (debug.enabled) debug('convertDecimalProps converted object: ', util.inspect(data));
+  } else {
+    const isDecimal = propDef && propDef.mongodb &&
+    propDef.mongodb.dataType &&
+    propDef.mongodb.dataType.toLowerCase() === 'decimal128';
     if (isDecimal) {
-      data[p] = Decimal128.fromString(data[p]);
-      debug('convertDecimalProps decimal value: ', data[p]);
+      data = Decimal128.fromString(data);
+      debug('convertDecimalProps decimal value: ', data);
```
b-admike: Let's add `if (debug.enabled)` condition here too for consistency.

test/decimal.test.js
```diff
@@ -8,59 +8,195 @@
...
-    OrderDecimal.create({count: '0.0005'}, function(err, order) {
-      if (err) return done(err);
-      // FIXME Ideally the returned data should have a decimal `count`
-      // not a string. While juggler generates the returned data, connector
-      // doesn't have any control.
-      order.count.should.equal('0.0005');
-      done();
+    it('destroyAll - deletes all with where', function(done) {
+      OrderDecimal.create({count: '0.0006'})
+        .then(function() {
```
b-admike: Can we create one more for `count: 0.0005`, so that our where condition for the delete is honoured?

test/decimal.test.js
```diff
@@ -8,59 +8,195 @@
...
+        // FIXME Ideally the returned data should have a decimal `count`
+        // not a string. While juggler generates the returned data, connector
+        // doesn't have any control.
+        JSON.parse(JSON.stringify(order.lines)).should.deepEqual(sample.lines);
+        done();
+      });
+    });
+
+    it('find - filters nested array decimal property', function(done) {
+      const cond = {where: {lines: {elemMatch: {unitPrice: Decimal128.fromString('0.0005')}}}};
```
b-admike: Maybe it might be beneficial to use `beforeEach` block to delete all records, and create/find/delete the records you want in each test. This removes hard dependencies between tests and test data pollution.

docs/decimal128.md
```diff
@@ -49,11 +88,28 @@ OrderDecimal.find({where: {count: '0.0005'}}, function(err, orders) {
 
 The connector automatically converts the condition from string to decimal.
 
+When query nested properties, you still need to do the conversion yourself, like
+
+```js
+// query decimal property in a nested array
+const arrCond = {where: {lines: {elemMatch: {unitPrice: Decimal128.fromString('0.0005')}}}};
```
bajtos: I have a question. How can LB users query nested Decimal128 properties via the REST API?

IIUC, the following query will not work because there is no code to automatically convert "0.0005" string into Decimal128 instance.

```ts
GET /orderdecimals?filter[where][lines][elemMatch][unitPrice]=0.0005
```

Is this something we should worry about and improve? (Possibly in a follow-up pull-request.)

jannyHou: @bajtos yes. Detecting the decimal property in a nested query is more complicated than in a model instance, since a query could contain extended operators. I would suggest customer use some workaround for the query...at least give the feature a lower severity level than sev1 to give us more time plan for it.

lib/mongodb.js
jannyHou: good catch.

test/decimal.test.js
jannyHou: @b-admike There are 3 models: OrderDecimal, OrderDecimalArr, OrderDecimalObj, so the 3 test suites won't have data pollution problem :)

And the 3 `it` tests in each `describe` block should run in order, it's on purpose (and please note I do clean up the data by running `destroyAll` in the before function):
- create a sample
- find the created sample with filter
- create another sample with different data --> destroy all data with filter --> find all data --> only one instance returned.
- find it by query

docs/decimal128.md
bajtos: @jannyHou Makes sense. Could you please open a follow-up GH issue for this?

PR #484, opened 2018-11-07
fix $where

jannyHou:
### Description

Fix test failure in https://github.com/strongloop/loopback-connector-mongodb/pull/483

According to [this juggler PR which introduced a flag to allow extended operators](https://github.com/strongloop/loopback-datasource-juggler/commit/a761e0d114486ea547ad0ae4391cfa3faccc9580#diff-d717a01e40d8164976e4468af1bce663R1845), the original test fails because `allowExtendedOperators` defaults to `false`, `Post.find()` throws "OPERATOR_NOT_ALLOWED_IN_QUERY" error instead of ignoring the nested `$where` filter.

I think the test was aimed at making sure the nested `$where` is not executed when  `allowExtendedOperators` is true, so set the flag to true in this fix PR. 

I don't know how the test passed before without throwing error. Without my change in the `buildWhere` function, it adds an additional `$` to the `$where`, resulting in mongodb driver throws "invalid operator $$where", which also makes sense. 

I am ok to either keep/remove my change in `buildWhere` function because they both prevents executing the nested `$where`. Opinions are welcomed :)

#### Related issues

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -951,6 +951,7 @@ MongoDB.prototype.buildWhere = function(model, where, options) {
       cond = cond[spec];
     }
     if (spec) {
+      if (spec.charAt(0) === '$') spec = spec.substr(1);
       if (spec === 'between') {
         query[k] = {$gte: cond[0], $lte: cond[1]};
       } else if (spec === 'inq') {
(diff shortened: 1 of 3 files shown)


lib/mongodb.js
```diff
@@ -990,6 +990,8 @@ MongoDB.prototype.buildWhere = function(model, where, options) {
           g.warn('{{MongoDB}} regex syntax does not respect the {{`g`}} flag');
 
         query[k] = {$regex: cond};
+      } else if (spec === '$where') {
+        query[k] = {$where: cond};
```
raymondfeng: We use `query[k]['$' + spec] = cond;` to map LB operators to mongodb. Maybe we should check the `$` prefix to avoid `$$`.

jannyHou: @raymondfeng I see good point.

3z3qu13l:
Hello guys,
Can we expect this to be merged and tagged in master soon ?

PR #498, opened 2019-02-25
test: remove decimal coercion from queries

b-admike:
### Description
Following the fix in `loopback-connector` to get nested property definitions in https://github.com/strongloop/loopback-connector/pull/136/ and https://github.com/strongloop/loopback-connector/pull/138, we need to modify tests calling find/delete to not coerce queries into decimal values. Before there was a limitation where we receive `undefined` for the property definition for nested properties, but now we can call these functions using string values as intended in the docs.

### Checklist

### DIFF
--- a/test/decimal.test.js
+++ b/test/decimal.test.js
@@ -169,7 +169,7 @@ describe('decimal128', function() {
     });
 
     it('find - filters nested object decimal property', function(done) {
-      const cond = {where: {'summary.totalValue': Decimal128.fromString('100.0005')}};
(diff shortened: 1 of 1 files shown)


b-admike:
Thanks for letting me know the expected behaviour @jannyHou 😄

b-admike:
@slnode test please

b-admike:
@jannyHou the tests on windows have not been passing for the past 4.5 months (thanks @rmg for the tip), and the changes are minor, so I'd like to ignore them given all other builds are passing.

PR #501, opened 2019-04-02
Fix/nested decimal props

### LINKED ISSUE #493: Loopback Issue with Update Operation - Data Type not stored correctly in Mongo DB
I am not able to update the document with correct field data type in Mongo DB using loopback mechanism. I tried with the model JSON file but it is only work for create operation but not for the update operation.

Loopback update operation defining the data types in a model file, is only working for the fields present at root level but not the fields present at Object, Arrary or nested object, nested array level. Fields present at root level alone are updated and saved correctly with the data type defined in model JSON file.

I have attached the sample model JSON file (movies.txt) along with the test output data (movies_sample_output.txt).  I have tested the behavior with the loopback-connector-mongodb@3.9.2 version.

I am new to loopback coding, so can someone help check the issue and implement the fix?  I can perform the testing and confirm the results. 

Please let me know for any more details needed. Thank you in advance..!!

[movies.txt](https://github.com/strongloop/loopback-connector-mongodb/files/2815318/movies.txt)
[Movies_sample_output.txt](https://github.com/strongloop/loopback-connector-mongodb/files/2815319/Movies_sample_output.txt)

b-admike:
### Description
Fixes #493; Coerce decimal property values from string to decimal128 at every level of the model definition tree. The decimal properties can be nested in an Array, Object, and any combination of those. Depends on https://github.com/strongloop/loopback-datasource-juggler/pull/1702 for model with property `randomReviews` (Array of decimal properties) to work.

#### Related issues

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -409,10 +409,11 @@ MongoDB.prototype.fromDatabase = function(model, data) {
  * @param {Object} data The JSON data to convert
  */
 MongoDB.prototype.toDatabase = function(model, data) {
+  const modelInstance = this._models[model].model;
   var props = this._models[model].properties;
 
   if (this.settings.enableGeoIndexing !== true) {
-    convertDecimalProps(data, props);
+    visitAllProperties(data, modelInstance, coerceDecimalProperty);
     // Override custom column names
     data = this.fromPropertyToDatabaseNames(model, data);
     return data;
@@ -429,7 +430,7 @@ MongoDB.prototype.toDatabase = function(model, data) {
     }
   }
 
-  convertDecimalProps(data, props);
+  visitAllProperties(data, modelInstance, coerceDecimalProperty);
   // Override custom column names
   data = this.fromPropertyToDatabaseNames(model, data);
   if (debug.enabled) debug('toDatabase data: ', util.inspect(data));
@@ -1996,23 +1997,75 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
 }
 
 /**
- * Convert the decimal properties from string to decimal.
- * Only supported after mongodb 3.4
  *
- * @param {Object} data The data that might contain a decimal property
- * @param {Object} props The model property definitions
+ * @param {*} data Plain Data Object for the matching property definition(s)
+ * @param {*} modelCtorOrDef Model constructor or definition
+ * @param {*} visitor A callback function which takes a property value and
+ * definition to apply custom property coercion
  */
-function convertDecimalProps(data, props) {
-  if (debug.enabled) debug('convertDecimalProps props: ', util.inspect(props));
-  for (const p in props) {
-    const prop = props[p];
-    const isDecimal = data[p] && prop && prop.mongodb &&
-      prop.mongodb.dataType &&
-      prop.mongodb.dataType.toLowerCase() === 'decimal128';
-    if (isDecimal) {
-      data[p] = Decimal128.fromString(data[p]);
-      debug('convertDecimalProps decimal value: ', data[p]);
+function visitAllProperties(data, modelCtorOrDef, visitor) {
+  if (data === null || data === undefined) return;
+  const modelProps = modelCtorOrDef.properties ? modelCtorOrDef.properties : modelCtorOrDef.definition.properties;
+  const allProps = new Set(Object.keys(data).concat(Object.keys(modelProps)));
+  for (const p of allProps) {
+    const value = data[p];
+    const def = modelProps[p];
+    if (def && def.type && isNestedModel(def.type)) {
+      if (Array.isArray(def.type)) {
+        if (value === null || value === undefined) break;
+        for (const it of value) {
+          visitAllProperties(it, def.type[0].definition, visitor);
+        }
+      } else {
+        if (value === null) break;
+        visitAllProperties(value, def.type.definition, visitor);
+      }
+    } else {
+      visitor(value, def, (newValue) => { data[p] = newValue; });
     }
   }
-  return data;
 }
+
+/**
+ *
+ * @param {*} propValue Property value to coerce into a Decimal128 value
+ * @param {*} propDef Property definition to check if property is MongoDB
+ * Decimal128 type
+ */
+function coerceDecimalProperty(propValue, propDef, setValue) {
+  let updatedValue;
+  if (hasDataType('decimal128', propDef)) {
+    if (Array.isArray(propValue)) {
+      updatedValue = propValue.map(val => Decimal128.fromString(val));
+      return setValue(updatedValue);
+    } else {
+      updatedValue = Decimal128.fromString(propValue);
+      return setValue(updatedValue);
+    }
+  }
+}
+
+/**
+* A utility function which checks for nested property definitions
+*
+* @param {*} propType Property type metadata
+*
+*/
+function isNestedModel(propType) {
+  if (!propType) return false;
+  if (Array.isArray(propType)) return isNestedModel(propType[0]);
+  return propType.definition && propType.definition.properties;
+}
+
+/**
+* A utility function which checks if a certain property definition matches
+* the given data type
+* @param {*} dataType The data type to check the property definition against
+* @param {*} propertyDef A property definition containing metadata about property type
+*/
+function hasDataType(dataType, propertyDef) {
+  return propertyDef && propertyDef.mongodb &&
+    propertyDef.mongodb.dataType &&
+    propertyDef.mongodb.dataType.toLowerCase() === dataType.toLowerCase();
+}
+
--- a/test/decimal.test.js
+++ b/test/decimal.test.js
@@ -6,6 +6,7 @@
 'use strict';
 
 require('./init.js');
+const promisify = require('bluebird').promisify;
 const Decimal128 = require('mongodb').Decimal128;
 
 var db, OrderDecimal;
@@ -48,8 +49,8 @@ describe('model with decimal property', function() {
     });
   });
 
-  it('destroyAll - deletes all with where', function(done) {
-    OrderDecimal.create({count: '0.0006'})
+  it('destroyAll - deletes all with where', function() {
+    return OrderDecimal.create({count: '0.0006'})
       .then(function() {
         return OrderDecimal.destroyAll({count: '0.0005'});
       })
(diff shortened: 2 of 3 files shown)


test/mongodb.test.js
```diff
@@ -547,6 +687,200 @@ describe('mongodb connector', function() {
...
+                  'testDecimal': '55.55',
+                },
+              },
+            },
+          },
+          'duration': '135',
+          'rating': '4.5',
+        },
+      }, function(err, inst) {
+        // inst.toObject().awards.prizeMoney.should.equal(Decimal128.fromString('25000.00'));
```
jannyHou: any reason to comment this assertion?

test/mongodb.test.js
```diff
@@ -547,6 +687,200 @@ describe('mongodb connector', function() {
...
+                },
+              },
+            },
+            'duration': '135',
+            'rating': '7.5',
+          },
+        };
+        modelWithDeepNestedDecimalProps.updateAll({id: inst.id}, updateData, function(err, inst) {
+          inst.toObject().imdb.rating.should.equal(Decimal128.fromString('7.5'));
+        });
```
jannyHou: do you want to test 'testDecimal' in the super nested `innerObj` property as well(since you defined it in the model)?

test/mongodb.test.js
b-admike: There is a bug in juggler right now where these assertions will fail because juggler doesn't return the updated model instance on create/update :(. Even though they are stored as decimal128 in mongo, we get back string values. I aim to fix that in juggler, but I was manually able to verify that they are stored correctly.

lib/mongodb.js
```diff
@@ -1996,23 +1996,97 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+      }
+    }
+  }
+
+  /**
+ * A utility function which checks if a certain property definition is
+ * decimal128 type
+ * @param {*} propertyDef A property definition containing metadata about property type
+ */
+  function checkDecimalProp(propertyDef) {
```
bajtos:
```suggestion
  function isDecimal128(propertyDef) {
```

Personally, I'd use a more generic variant.

```ts
function hasDataType(dataType, propertyDef) {
  return propertyDef && propertyDef.mongodb &&
    propertyDef.mongodb.dataType &&
    propertyDef.mongodb.dataType.toLowerCase() == dataType.toLowerCase();
}

// usage:
if (hasDataType('decimal128', propertyDef) {
  // coerce the properties
}
```

lib/mongodb.js
```diff
@@ -1996,23 +1996,97 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+      if (!propDefinition.type[0].definition) {
+        return false;
+      } else {
+        if (propDefinition.type[0].definition.properties) return true;
+      }
+    } else {
+      if (!propDefinition.type.definition) {
+        return false;
+      } else {
+        if (propDefinition.type.definition.properties) return true;
```
bajtos: I think this function can be simplified by accepting the `type` instead of full property definition.

```js
function isNestedModel(propType) {
  if (!propType) return false;
  if (Array.isArray(propType)) return isNestedModel(propType[0]);
  return propType.definition && propType.definition.properties;
}
```

test/mongodb.test.js
```diff
@@ -308,6 +312,142 @@ describe('mongodb connector', function() {
       }
     );
 
+    modelWithDecimalArray = db.define('modelWithDecimalArray', {
```
bajtos: AFAICT, each of these models are used by a single test only. Please move their definitions to the tests that are using them, it will make the tests easier to read and also will avoid temptation to add unrelated properties to these models in the future.

lib/mongodb.js
```diff
@@ -1996,23 +1996,97 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+  visitAllProperties(data, props, coerceDecimalProperties);
+
+  /**
+ *
+ * @param {*} data Plain Data Object for the matching property definition(s)
+ * @param {*} propDefMap Property definition(s) which include information about property type
+ * @param {*} visitor A callback function which takes a property value and
+ * definition to apply custom property coercion
+ */
+  function visitAllProperties(data, propDefMap, visitor) {
```
bajtos: IMPORTANT❗️ 

IIUC, these new functions are defined inside `convertDecimalProps`. Since they are not (or at least should not) require anything from `convertDecimalProps` closure, they should be defined at the top-level, so that there is only one ("singleton") instance used by all invocations of `convertDecimalProps`. Such solution allows V8 to optimize performance.

Personally, I'd prefer to move this particular helper to `loopback-connector`, so that connectors for other databases can use it too. To land this pull request sooner, I think it will be best to make that change (move) after this patch is landed.

lib/mongodb.js
```diff
@@ -1996,23 +1996,97 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+    }
+  }
+
+  /**
+   *
+   * @param {*} value Property value to coerce into a Decimal128 value
+   * @param {*} def Property definition to check if property is MongoDB
+   * decima type
+   */
+  function coerceDecimalProperties(value, def) {
```
bajtos: Ditto, move this function out of `convertDecimalProps`.

lib/mongodb.js
```diff
@@ -1996,23 +1996,97 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
     }
   }
+
+  /**
+ * A utility function which checks for nested property definitions
+ *
+ * @param {*} propDefinition Property definition which contains metadata for the
+ * property type
+ */
+  function checkNested(propDefinition) {
```
bajtos: Ditto, move this function out of `convertDecimalProps`.

lib/mongodb.js
bajtos: Ditto, move this function out of `convertDecimalProps`.

test/mongodb.test.js
bajtos: > because juggler doesn't return the updated model instance on create/update

I think that's an intentional behavior, see https://loopback.io/doc/en/lb3/Operation-hooks.html

> By default, create and updateAttributes do not apply database updates to the model instance returned to the callback, therefore any changes made by “loaded” hooks are discarded. To change this behavior, set a per-model option `updateOnLoad: true`.

I think it's more important to verify what has been actually stored by the database. You can start by calling `findById` to fetch the actual data. IDK if `decimal128` type is preserved on that path. If not, then you may need to execute custom MongoDB command to access raw data.

test/mongodb.test.js
```diff
@@ -547,6 +687,200 @@ describe('mongodb connector', function() {
...
+  context('nested decimal props', function() {
+    it('should create/update instance for array of decimal props', function(done) {
+      modelWithDecimalArray.create({
+        'randomReview': [
+          '3.5',
+          '4.5',
+          '4.0',
+        ],
+      }, function(err, inst) {
+        inst.toObject().randomReview[0].should.equal(Decimal128.fromString('3.5'));
```
bajtos: I am little bit confused about this assertion. Are you sure that `equal` check is verifying whether both values are instances of `Decimal128`? Shouldn't you use `instanceof` operator instead?

I guess the more important question is whether you have seen these new tests fail on this assertion and what were the error messages reported?

lib/mongodb.js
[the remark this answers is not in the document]
b-admike: I like the proposal for the latter where we can preserve `visitAllProperties` but call `visitor` as `setValue` instead 👍

lib/mongodb.js
```diff
@@ -1996,23 +1996,97 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
-      prop.mongodb.dataType.toLowerCase() === 'decimal128';
-    if (isDecimal) {
-      data[p] = Decimal128.fromString(data[p]);
-      debug('convertDecimalProps decimal value: ', data[p]);
+  visitAllProperties(data, props, coerceDecimalProperties);
+
+  /**
+ *
+ * @param {*} data Plain Data Object for the matching property definition(s)
+ * @param {*} propDefMap Property definition(s) which include information about property type
```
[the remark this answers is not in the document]
b-admike: I have not but can take a stab at it.

lib/mongodb.js
b-admike: That's an elegant way to put it, thank you I agree.

lib/mongodb.js
b-admike: >IIUC, these new functions are defined inside convertDecimalProps. Since they are not (or at least should not) require anything from convertDecimalProps closure, they should be defined at the top-level, so that there is only one ("singleton") instance used by all invocations of convertDecimalProps. Such solution allows V8 to optimize performance.

Good to know, thanks! That makes a lot of sense. Yea we can think of moving this function out to `loopback-connector` as discussed after this patch is landed.

test/mongodb.test.js
b-admike: Yes the tests fail like so:
```
  1) mongodb connector
       nested decimal props
         should create/update instance for array of decimal props:
     Uncaught AssertionError: expected '3.5' to be Decimal128 {
  _bsontype: 'Decimal128',
  bytes: Buffer [ 23, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 3e, 30 ]
}
```
I think using `instanceOf` operator is better, but like you mentioned above, I might have to use a custom `execute` call to grab the raw data from mongodb.

test/mongodb.test.js
b-admike: `findById` did not work, but yeah perhaps using custom MongoDB command would work here 👍

test/mongodb.test.js
bajtos: > ```
>  Uncaught AssertionError: expected '3.5' to be Decimal128 {
>  _bsontype: 'Decimal128',
>  bytes: Buffer [ 23, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 3e, 30 ]
> }
> ```

Nice!

lib/mongodb.js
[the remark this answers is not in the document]
bajtos: I am afraid you misunderstood my second proposal. I was not saying to rename `visitor` to `setValue`, that would be confusing.

I'll post a new comment in the updated code.

lib/mongodb.js
```diff
@@ -1996,23 +1997,74 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
-  if (debug.enabled) debug('convertDecimalProps props: ', util.inspect(props));
-  for (const p in props) {
-    const prop = props[p];
-    const isDecimal = data[p] && prop && prop.mongodb &&
-      prop.mongodb.dataType &&
-      prop.mongodb.dataType.toLowerCase() === 'decimal128';
-    if (isDecimal) {
-      data[p] = Decimal128.fromString(data[p]);
-      debug('convertDecimalProps decimal value: ', data[p]);
+function visitAllProperties(data, modelCtorOrDef, setValue) {
```
bajtos: Continuing discussion started earlier.

A function called `visitAllProperties` should not be modifying property values, it should only call the supplied `visiter` function for every property.

```suggestion
function visitAllProperties(data, modelCtorOrDef, visitor) {
```

lib/mongodb.js
```diff
@@ -1996,23 +1997,74 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+        if (value === null || value.length === 0) continue;
+        for (const it of value) {
+          visitAllProperties(it, def.type[0].definition, setValue);
+        }
+      } else {
+        if (value === null) continue;
+        visitAllProperties(value, def.type.definition, setValue);
+      }
+    } else {
+      data[p] = setValue(value, def);
```
bajtos: Continuing discussion started earlier. Instead of using the value returned by the visitor to update the property, we should provide the visitor function with a tool allowing the visitor to decide if the property value should be updated or not.

```suggestion
      visitor(value, def, (newValue) => { data[p] = newValue; });
```

lib/mongodb.js
```diff
@@ -1996,23 +1997,74 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
-  return data;
 }
+
+/**
+ *
+ * @param {*} propValue Property value to coerce into a Decimal128 value
+ * @param {*} propDef Property definition to check if property is MongoDB
+ * decima type
+ */
+function coerceDecimalProperties(propValue, propDef) {
```
bajtos: Since this is coercing a single property, I think a singular form would be better?

```suggestion
function coerceDecimalProperty(propValue, propDef) {
```

lib/mongodb.js
```diff
@@ -1996,23 +1997,74 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
     }
   }
-  return data;
 }
+
+/**
+ *
+ * @param {*} propValue Property value to coerce into a Decimal128 value
+ * @param {*} propDef Property definition to check if property is MongoDB
+ * decima type
```
bajtos: Typo: decima -> decimal.

Personally, I would use the exact type name as used by MongoDB, i.e. `decimal128`.

```suggestion
 * decimal128 type
```

lib/mongodb.js
```diff
@@ -1996,23 +1997,74 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+
+/**
+* A utility function which checks if a certain property definition matches
+* the given data type
+* @param {*} dataType The data type to check the property definition against
+* @param {*} propertyDef A property definition containing metadata about property type
+*/
+function hasDataType(dataType, propertyDef) {
+  return propertyDef && propertyDef.mongodb &&
+propertyDef.mongodb.dataType &&
```
bajtos:
```suggestion
    propertyDef.mongodb.dataType &&
```

lib/mongodb.js
```diff
@@ -1996,23 +1997,74 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
+propertyDef.mongodb.dataType.toLowerCase() === dataType.toLowerCase();
```
bajtos:
```suggestion
    propertyDef.mongodb.dataType.toLowerCase() === dataType.toLowerCase();
```

test/decimal.test.js
```diff
@@ -63,4 +69,431 @@ describe('model with decimal property', function() {
...
+          createdInstance.randomReview[0].should.deepEqual(Decimal128.fromString('3.5'));
+          return modelWithDecimalArray.updateAll({id: instanceId}, updateData);
+        })
+        .then(function(inst) {
+          return findModelInstanceAsync('modelWithDecimalArray', instanceId);
+        })
+        .then(function(updatedInstance) {
+          updatedInstance.randomReview[0].should.be.instanceOf(Decimal128);
+          updatedInstance.randomReview[0].should.deepEqual(Decimal128.fromString('5.5'));
+          done();
```
bajtos: Don't mix callbacks with promises. Mocha supports tests written with promises, use that feature please.

```ts
it('should create/update instance for array of decimal props', function() {
  // ...
  return modelWithDecimalArray.create(createData)
    // ...
   .then(function(updatedInstance) {
      updatedInstance.randomReview[0].should.be.instanceOf(Decimal128);
      updatedInstance.randomReview[0].should.deepEqual(Decimal128.fromString('5.5'));
    });
});
```

The same comment applies to other tests too.

test/decimal.test.js
```diff
@@ -63,4 +69,431 @@ describe('model with decimal property', function() {
...
+            .should.be.instanceOf(Decimal128);
+          updatedInstance.imdb.innerArray[0].nestedArray[0]
+            .testDecimal.should.be.instanceOf(Decimal128);
+          updatedInstance.imdb.innerArray[0].nestedArray[0]
+            .testDecimal.should.deepEqual(Decimal128.fromString('77.77'));
+          done();
+        });
+    });
+
+    function findModelInstance(modelName, id, cb) {
```
bajtos: IIUC, this method is not returning an instance of the model class, just the raw model data. In which case I am proposing a slightly different name:

```suggestion
    function findRawModelData(modelName, id, cb) {
```

Similarly for `findModelInstanceAsync` below.

test/decimal.test.js
```diff
@@ -63,4 +69,431 @@ describe('model with decimal property', function() {
...
+      db.connector.execute(modelName, 'findOne', {_id: {$eq: id}}, {safe: true}, cb);
+    }
+    var findModelInstanceAsync = promisify(findModelInstance);
```
bajtos: Use `const`, not `var` - see https://loopback.io/doc/en/contrib/style-guide-es6.html#variable-declarations

lib/mongodb.js
b-admike: Agreed, this had unintended behaviour (backwards-incompatible) that was setting undefined values in the data object when we should skip them. Thank you!

b-admike:
For the last commit, I used https://github.com/strongloop/loopback-connector-mongodb/pull/483/files#diff-f44d7a6206dcaa5ddd05b31118b5c2be as inspiration to fix one of the failing tests on master.

b-admike:
Thanks for all the feedback @bajtos. I've applied them all. LGTY now?

test/mongodb.test.js
```diff
@@ -345,7 +345,7 @@ describe('mongodb connector', function() {
       });
       ds.ping(function(err) {
         (!!err).should.be.True();
-        err.message.should.match(/failed to connect to server/);
+        err.code.should.eql('ECONNREFUSED');
```
bajtos: Please move `ECONNREFUSED ` assertion changes to a standalone commit, or even to a standalone pull request (to get them landed quickly).

test/decimal.test.js
```diff
@@ -59,8 +65,426 @@ describe('model with decimal property', function() {
...
+            'innerObj': {
+              'innerObj': {
+                'innerObj': {
+                  'testNumber': '6666',
+                  'testDate': '2019-01-30 18:26:38.551022',
+                  'testDecimal': '55.55',
+                },
+              },
+            },
+          },
```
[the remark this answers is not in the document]
b-admike: That sounds good. I will make a follow-up PR for these improvements 👍

b-admike:
The test failures below are what #505 aims to fix and are not introduced by this PR (also fail on master):
```
10:08:59   1128 passing (17s)
10:08:59   100 pending
10:08:59 
  3 failing
10:08:59 
10:08:59   1) lazyConnect
10:08:59        should report connection error (lazyConnect = false):
10:08:59      Uncaught AssertionError: expected 'connect ECONNREFUSED 127.0.0.1:4' to match /failed to connect to server/
10:08:59       at Assertion.fail (node_modules/should/cjs/should.js:275:17)
10:08:59       at Assertion.value [as match] (node_modules/should/cjs/should.js:356:19)
10:08:59       at DataSource.<anonymous> (test/mongodb.test.js:59:26)
10:08:59       at DataSource.postInit (node_modules/loopback-datasource-juggler/lib/datasource.js:480:16)
10:08:59       at /home/jenkins/workspace/nb/loopback-connector-mongodb~master/3ed5ebba/lib/mongodb.js:320:23
10:08:59       at err (node_modules/mongodb/lib/utils.js:411:14)
10:08:59       at executeCallback (node_modules/mongodb/lib/utils.js:401:25)
10:08:59       at err (node_modules/mongodb/lib/operations/mongo_client_ops.js:285:21)
10:08:59       at connectCallback (node_modules/mongodb/lib/operations/mongo_client_ops.js:265:5)
10:08:59       at server.connect (node_modules/mongodb/lib/operations/mongo_client_ops.js:353:14)
10:08:59       at Server.<anonymous> (node_modules/mongodb/lib/topologies/server.js:237:11)
10:08:59       at Pool.<anonymous> (node_modules/mongodb-core/lib/topologies/server.js:452:12)
10:08:59       at connect (node_modules/mongodb-core/lib/connection/pool.js:557:14)
10:08:59       at makeConnection (node_modules/mongodb-core/lib/connection/connect.js:38:11)
10:08:59       at callback (node_modules/mongodb-core/lib/connection/connect.js:260:5)
10:08:59       at Socket.err (node_modules/mongodb-core/lib/connection/connect.js:286:7)
10:08:59       at emitErrorNT (internal/streams/destroy.js:66:8)
10:08:59       at _combinedTickCallback (internal/process/next_tick.js:139:11)
10:08:59       at process._tickCallback (internal/process/next_tick.js:181:9)
10:08:59 
10:08:59   2) mongodb connector
10:08:59        does not execute a nested `$where`:
10:08:59      Uncaught AssertionError: expected Error {
10:08:59   code: 'OPERATOR_NOT_ALLOWED_IN_QUERY',
10:08:59   statusCode: 400,
10:08:59   details: Object {
10:08:59     operators: Array [ '$where' ],
10:08:59     where: Object {
10:08:59       content: Object {
10:08:59         $where: 'function() {return this.content.contains("content")}'
10:08:59       }
10:08:59     }
10:08:59   },
10:08:59   message: 'Operators "$where" are not allowed in query'
10:08:59 } to not exist
10:08:59       at Post.find (test/mongodb.test.js:830:24)
10:08:59       at /home/jenkins/workspace/nb/loopback-connector-mongodb~master/3ed5ebba/node_modules/loopback-datasource-juggler/lib/dao.js:1532:7
10:08:59       at _combinedTickCallback (internal/process/next_tick.js:132:7)
10:08:59       at process._tickCallback (internal/process/next_tick.js:181:9)
10:08:59 
10:08:59   3) mongodb connector
10:08:59        .ping(cb)
10:08:59          should report connection errors with invalid config:
10:08:59      Uncaught AssertionError: expected 'connect ECONNREFUSED 127.0.0.1:4' to match /failed to connect to server/
10:08:59       at Assertion.fail (node_modules/should/cjs/should.js:275:17)
10:08:59       at Assertion.value [as match] (node_modules/should/cjs/should.js:356:19)
10:08:59       at /home/jenkins/workspace/nb/loopback-connector-mongodb~master/3ed5ebba/test/mongodb.test.js:348:28
10:08:59       at DataSource.<anonymous> (lib/mongodb.js:1863:7)
10:08:59       at DataSource.postInit (node_modules/loopback-datasource-juggler/lib/datasource.js:480:16)
10:08:59       at /home/jenkins/workspace/nb/loopback-connector-mongodb~master/3ed5ebba/lib/mongodb.js:320:23
10:08:59       at err (node_modules/mongodb/lib/utils.js:411:14)
10:08:59       at executeCallback (node_modules/mongodb/lib/utils.js:401:25)
10:08:59       at err (node_modules/mongodb/lib/operations/mongo_client_ops.js:285:21)
10:08:59       at connectCallback (node_modules/mongodb/lib/operations/mongo_client_ops.js:265:5)
10:08:59       at server.connect (node_modules/mongodb/lib/operations/mongo_client_ops.js:353:14)
10:08:59       at Server.<anonymous> (node_modules/mongodb/lib/topologies/server.js:237:11)
10:08:59       at Pool.<anonymous> (node_modules/mongodb-core/lib/topologies/server.js:452:12)
10:08:59       at connect (node_modules/mongodb-core/lib/connection/pool.js:557:14)
10:08:59       at makeConnection (node_modules/mongodb-core/lib/connection/connect.js:38:11)
10:08:59       at callback (node_modules/mongodb-core/lib/connection/connect.js:260:5)
10:08:59       at Socket.err (node_modules/mongodb-core/lib/connection/connect.js:286:7)
10:08:59       at emitErrorNT (internal/streams/destroy.js:66:8)
10:08:59       at _combinedTickCallback (internal/process/next_tick.js:139:11)
10:08:59       at process._tickCallback (internal/process/next_tick.js:181:9)
10:08:59 
```
Therefore, I'm going to merge this PR and hopefully we can fix them in #505.

Issue #503, opened 2019-04-06
Fix CI on OPERATOR_NOT_ALLOWED_IN_QUERY


dhmlau:
## Description
Currently there is one test case failed in CI because of the error below:

```
mongodb connector
does not execute a nested $where:
Uncaught AssertionError: expected Error {
code: 'OPERATOR_NOT_ALLOWED_IN_QUERY',
statusCode: 400,
details: Object {
operators: Array [ '$where' ],
where: Object {
content: Object {
$where: 'function() {return this.content.contains("content")}'
}
}
},
message: 'Operators "$where" are not allowed in query'
} to not exist
at Post.find (test/mongodb.test.js:830:24)
at /home/travis/build/strongloop/loopback-connector-mongodb/node_modules/loopback-datasource-juggler/lib/dao.js:1525:7
at process._tickCallback (internal/process/next_tick.js:61:11)

```

_Extracted from https://github.com/strongloop/loopback-connector-mongodb/pull/497_

There is a setting in mongodb, there is a setting : allow extended operators 
In juggler,  there are security settings that may have contributed to this failure (introduced ~ 3 months ago)

Related to this change : https://github.com/strongloop/loopback-connector-mongodb/pull/483/files#diff-f44d7a6206dcaa5ddd05b31118b5c2beR786 

## Acceptance Criteria
- [ ] Fix CI and make it all green again
  - [ ] `master` branch
  - [ ] `3.x` branch

dhmlau:
Also see https://github.com/strongloop/loopback-datasource-juggler/pull/1705#issuecomment-480254509

dhmlau:
See PR #505

b-admike:
Fixed by #505, created #509 for follow up removal of downstream builds.

PR #507, opened 2019-04-19
Keep undefined embedded model properties instead of converting to null

### LINKED ISSUE #506: undefined converted to null when missing from db

When a property has a ModelDefinition as type declared, it is recursively passed from `fromDatabase` to be converted in case  of an embedded Binary property. You can see this in this [line](https://github.com/strongloop/loopback-connector-mongodb/blob/cf10c6be5d6a00a95dc00bfeaf10ec260cd94e60/lib/mongodb.js#L398).

However when an empty (undefined) property is met, it calls the fn recursively and returns `null` as value, which results in the value changed from `undefined` to `null` in the model. You can see this [here](https://github.com/strongloop/loopback-connector-mongodb/blob/cf10c6be5d6a00a95dc00bfeaf10ec260cd94e60/lib/mongodb.js#L371)

**For example**
A model with definition of
```js
// ModelA
{
  name: 'string',
  location: 'Location'
}
```
where `Location` is another model:
```js
// Location
{
  lat: 'number',
  lng: 'number'
}
```
and a db entry exists with data: `{ name: 'someName' }`, when retrieved it will be converted to
`{ name: 'someName', location: null }` which is inconsistent with the existing data

# Expected result
It should actually keep the actual data type and not add null properties when undefined, which is already happening for simple properties.

# Additional information
linux x64 10.15.3
├── loopback@3.25.1
├── loopback-connector-mongodb@4.1.0

mitsos1os:
### Description
fixes #506 

Add extra check for existing value before recursively trying to convert to database type

#### Related issues

- connect to #506 

### Checklist

### DIFF
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -29,6 +29,7 @@ var Superhero,
   PostWithStringIdAndRenamedColumns,
   Employee,
   PostWithDisableDefaultSort,
+  WithEmbeddedProperties,
   WithEmbeddedBinaryProperties;
 
 describe('lazyConnect', function() {
@@ -295,6 +296,20 @@ describe('mongodb connector', function() {
       }
     );
 
+    WithEmbeddedProperties = db.define(
+      'WithEmbeddedProperties',
+      {
+        id: {type: String, id: true},
+        name: {type: String},
+        location: {
+          type: {
+            city: {type: String},
+            country: {type: String},
+          },
+        },
+      }
+    );
+
     WithEmbeddedBinaryProperties = db.define(
       'WithEmbeddedBinaryProperties',
       {
@@ -322,7 +337,9 @@ describe('mongodb connector', function() {
               PostWithStringId.destroyAll(function() {
                 PostWithDisableDefaultSort.destroyAll(function() {
                   Category.destroyAll(function() {
-                    done();
+                    WithEmbeddedProperties.destroyAll(function() {
+                      done();
+                    });
                   });
                 });
               });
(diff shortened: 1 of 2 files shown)


slnode:

dhmlau:
@slnode test please

mitsos1os:
@dhmlau Thank you for taking the time to review it!

test/mongodb.test.js
```diff
@@ -701,6 +718,34 @@ describe('mongodb connector', function() {
...
 
+  it('should properly retrieve embedded model properties', function(done) {
+    const data = {name: 'Mitsos', location: {city: 'Volos', country: 'Greece'}};
+    WithEmbeddedProperties.create(data, function(err, createdModel) {
+      if (err) return done(err);
+      WithEmbeddedProperties.findById(createdModel.id, function(err, dbModel) {
+        if (err) return done(err);
+        const modelObj = dbModel.toJSON();
+        const dataObj = Object.assign({id: modelObj.id}, data);
+        modelObj.should.be.eql(dataObj);
```
b-admike: Can we use `deepEqual` for object property values here i.e. `modelObj.should.deepEqual(dataObj);`

mitsos1os: `eql` is the same with `deepEqual` and is also used in more places in the file. Would you like to change it to the `deepEqual` you suggest ?

b-admike: In that case, fine by me to keep it as `eql` 👍

mitsos1os: OK cool! Thanks! I am resolving this so the PR moves on

dhmlau:
CI for downstream is failing. we have a ticket opened to fix the CI https://github.com/strongloop/loopback-connector-mongodb/issues/503.

dhmlau:
@mitsos1os, thanks for your contribution.  Your PR has landed! 🎉

PR #513, opened 2019-04-30
Fix/nested decimal coercion logic

b-admike:
### Description
Refactor nested decimal type tests from https://github.com/strongloop/loopback-connector-mongodb/pull/501#discussion_r274368255 and also fix attempt to convert undefined/null values for decimal properties.

#### Related issues

Related to #493 

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -2013,19 +2013,19 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
   for (const p of allProps) {
     const value = data[p];
     const def = modelProps[p];
+    if (!value) continue;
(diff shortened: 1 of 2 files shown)


lib/mongodb.js
```diff
@@ -2015,17 +2015,18 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
     const def = modelProps[p];
     if (def && def.type && isNestedModel(def.type)) {
```
dhmlau: should we check with `value` is not null in this line instead of in line 2018 and 2026?

b-admike: Good point. Fixed.

b-admike:
@slnode test please

PR #517, opened 2019-05-09
feat: adds mongodb: {dataType: 'ObjectID'} to model properties

hacksparrow:
Adds `mongodb: {dataType: 'ObjectID'}` to model properties. Enforces `strictObjectIDCoercion` settings on all model properties.

#### Related issues

https://github.com/strongloop/loopback-next/issues/2085

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -18,6 +18,9 @@ const Connector = require('loopback-connector').Connector;
 const debug = require('debug')('loopback:connector:mongodb');
 const Decimal128 = mongodb.Decimal128;
 
+const ObjectIdValueRegex = /^[0-9a-fA-F]{24}$/;
+const ObjectIdTypeRegex = /objectid/i;
+
 exports.ObjectID = ObjectID;
 /*!
  * Convert the id to be a BSON ObjectID if it is compatible
@@ -35,7 +38,7 @@ function ObjectID(id) {
     // MongoDB's ObjectID constructor accepts number, 12-byte string or 24-byte
     // hex string. For LoopBack, we only allow 24-byte hex string, but 12-byte
     // string such as 'line-by-line' should be kept as string
-    if (/^[0-9a-fA-F]{24}$/.test(id)) {
+    if (ObjectIdValueRegex.test(id)) {
       return bson.ObjectID(id);
     } else {
       return id;
@@ -180,6 +183,7 @@ function MongoDB(settings, dataSource) {
   }
 
   this.dataSource = dataSource;
+
   if (
     this.settings.enableOptimisedfindOrCreate === true ||
     this.settings.enableOptimisedFindOrCreate === true ||
@@ -412,11 +416,11 @@ MongoDB.prototype.fromDatabase = function(modelName, data) {
  * @param {Object} data The JSON data to convert
  */
 MongoDB.prototype.toDatabase = function(modelName, data) {
-  const modelInstance = this._models[modelName].model;
-  const props = this._models[modelName].properties;
+  const modelCtor = this._models[modelName];
+  const props = modelCtor.properties;
 
   if (this.settings.enableGeoIndexing !== true) {
-    visitAllProperties(data, modelInstance, coerceDecimalProperty);
+    visitAllProperties(data, modelCtor, coercePropertyValue);
     // Override custom column names
     data = this.fromPropertyToDatabaseNames(modelName, data);
     return data;
@@ -433,7 +437,7 @@ MongoDB.prototype.toDatabase = function(modelName, data) {
     }
   }
 
-  visitAllProperties(data, modelInstance, coerceDecimalProperty);
+  visitAllProperties(data, modelCtor, coercePropertyValue);
   // Override custom column names
   data = this.fromPropertyToDatabaseNames(modelName, data);
   if (debug.enabled) debug('toDatabase data: ', util.inspect(data));
@@ -528,6 +532,7 @@ MongoDB.prototype.coerceId = function(modelName, id, options) {
 
   // Type conversion for id
   const idProp = self.getPropertyDefinition(modelName, idName);
+
   if (idProp && typeof idProp.type === 'function') {
     if (!(idValue instanceof idProp.type)) {
       idValue = idProp.type(id);
@@ -537,9 +542,8 @@ MongoDB.prototype.coerceId = function(modelName, id, options) {
       }
     }
 
-    if (self.isObjectIDProperty(modelName, idProp, idValue, options)) {
-      idValue = ObjectID(idValue);
-    }
+    const modelCtor = this._models[modelName];
+    idValue = coerceToObjectId(modelCtor, idProp, idValue);
   }
   return idValue;
 };
@@ -841,7 +845,7 @@ MongoDB.prototype.updateOrCreate = function updateOrCreate(
 /**
  * Replace model instance if it exists or create a new one if it doesn't
  *
- * @param {String} modelName The name of the model
+ * @param {String} modelName The model name
  * @param {Object} data The model instance data
  * @param {Object} options The options object
  * @param {Function} [cb] The callback function
@@ -942,12 +946,12 @@ MongoDB.prototype.buildWhere = function(modelName, where, options) {
 
     const prop = self.getPropertyDefinition(modelName, propName);
 
-    const isDecimal = prop &&
-      prop.mongodb && prop.mongodb.dataType &&
-      prop.mongodb.dataType.toLowerCase() === 'decimal128';
-    if (isDecimal) {
-      cond = Decimal128.fromString(cond);
-      debug('buildWhere decimal value: %s, constructor name: %s', cond, cond.constructor.name);
+    if (prop && prop.mongodb && typeof prop.mongodb.dataType === 'string') {
+      const isDecimal = prop.mongodb.dataType.toLowerCase() === 'decimal128';
+      if (isDecimal) {
+        cond = Decimal128.fromString(cond);
+        debug('buildWhere decimal value: %s, constructor name: %s', cond, cond.constructor.name);
+      }
     }
 
     // Convert property to database column name
@@ -960,6 +964,9 @@ MongoDB.prototype.buildWhere = function(modelName, where, options) {
       spec = Object.keys(cond)[0];
       cond = cond[spec];
     }
+
+    const modelCtor = self._models[modelName];
+
     if (spec) {
       if (spec === 'between') {
         query[k] = {$gte: cond[0], $lte: cond[1]};
@@ -967,7 +974,7 @@ MongoDB.prototype.buildWhere = function(modelName, where, options) {
         cond = [].concat(cond || []);
         query[k] = {
           $in: cond.map(function(x) {
-            if (self.isObjectIDProperty(modelName, prop, x, options))
+            if (isObjectIDProperty(modelCtor, prop, x, options))
               return ObjectID(x);
             return x;
           }),
@@ -976,7 +983,7 @@ MongoDB.prototype.buildWhere = function(modelName, where, options) {
         cond = [].concat(cond || []);
         query[k] = {
           $nin: cond.map(function(x) {
-            if (self.isObjectIDProperty(modelName, prop, x, options))
+            if (isObjectIDProperty(modelCtor, prop, x, options))
               return ObjectID(x);
             return x;
           }),
@@ -1010,7 +1017,7 @@ MongoDB.prototype.buildWhere = function(modelName, where, options) {
         // Null: 10
         query[k] = {$type: 10};
       } else {
-        if (self.isObjectIDProperty(modelName, prop, cond, options)) {
+        if (isObjectIDProperty(modelCtor, prop, cond, options)) {
           cond = ObjectID(cond);
         }
         query[k] = cond;
@@ -1217,8 +1224,6 @@ MongoDB.prototype.getDatabaseColumnName = function(model, propName) {
 
   const prop = model.properties[propName] || {};
 
-  // console.log('getDatabaseColumnName', propName, prop);
-
   // Try mongo overrides
   if (prop.mongodb) {
     propName =
@@ -1234,8 +1239,6 @@ MongoDB.prototype.getDatabaseColumnName = function(model, propName) {
     propName = prop.columnName || prop.column || propName;
   }
 
-  // Done
-  // console.log('->', propName);
   return propName;
 };
 
@@ -1870,30 +1873,72 @@ MongoDB.prototype.ping = function(cb) {
   }
 };
 
+// Case insensitive check if a string looks like "ObjectID"
+function typeIsObjectId(input) {
+  if (!input) return false;
+  return typeof input === 'string' && input.match(ObjectIdTypeRegex);
+}
+
+// Determine if a property must be stored as ObjectID
+function isStoredAsObjectID(propDef) {
+  if (!propDef) return false;
+
+  if (propDef.mongodb) {
+    if (ObjectIdTypeRegex.test(propDef.mongodb.dataType)) return true;
+  } else if (propDef.type) {
+    if (typeof propDef.type === 'string' && typeIsObjectId(propDef.type)) return true;
+    else if (Array.isArray(propDef.type)) {
+      if (propDef.type[0] === ObjectID || typeIsObjectId(propDef.type[0])) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+// Determine if strictObjectIDCoercion should be enabled
+function isStrictObjectIDCoercionEnabled(modelCtor, options) {
+  const settings = modelCtor.settings;
+  return (settings && settings.strictObjectIDCoercion) ||
+    (modelCtor.model && modelCtor.model.getConnector().settings.strictObjectIDCoercion) ||
+    options &&
+    options.strictObjectIDCoercion;
+}
+
+// Tries to coerce a property into ObjectID after checking multiple conditions
+function coerceToObjectId(modelCtor, propDef, propValue) {
+  if (isStoredAsObjectID(propDef)) {
+    if (isObjectIDProperty(modelCtor, propDef, propValue)) {
+      return ObjectID(propValue);
+    } else {
+      throw new Error(`${propValue} is not an ObjectID string`);
+    }
+  } else if (isStrictObjectIDCoercionEnabled(modelCtor)) {
+    if (isObjectIDProperty(modelCtor, propDef, propValue)) {
+      return ObjectID(propValue);
+    }
+  } else if (ObjectIdValueRegex.test(propValue)) {
+    return ObjectID(propValue);
+  }
+  return propValue;
+}
+
 /**
  * Check whether the property is an ObjectID (or Array thereof)
  *
  */
-MongoDB.prototype.isObjectIDProperty = function(modelName, prop, value, options) {
-  if (
-    prop &&
-    (prop.type === ObjectID ||
-      (Array.isArray(prop.type) && prop.type[0] === ObjectID))
-  ) {
+function isObjectIDProperty(modelCtor, propDef, value, options) {
+  if (!propDef) return false;
+
+  if (typeof value === 'string' && value.match(ObjectIdValueRegex)) {
+    if (isStoredAsObjectID(propDef)) return true;
+    else return !isStrictObjectIDCoercionEnabled(modelCtor, options);
+  } else if (value instanceof ObjectID) {
     return true;
-  } else if ('string' === typeof value) {
-    const settings = this._models[modelName] && this._models[modelName].settings;
-    options = options || {};
-    const strict =
-      (settings && settings.strictObjectIDCoercion) ||
-      this.settings.strictObjectIDCoercion ||
-      options.strictObjectIDCoercion;
-    if (strict) return false; // unless explicitly typed, don't coerce
-    return /^[0-9a-fA-F]{24}$/.test(value);
   } else {
     return false;
   }
-};
+}
 
 function sanitizeFilter(filter, options) {
   options = Object.assign({}, options);
@@ -2001,15 +2046,14 @@ function optimizedFindOrCreate(modelName, filter, data, options, callback) {
 }
 
 /**
- *
  * @param {*} data Plain Data Object for the matching property definition(s)
- * @param {*} modelCtorOrDef Model constructor or definition
+ * @param {*} modelCtor Model constructor
  * @param {*} visitor A callback function which takes a property value and
  * definition to apply custom property coercion
  */
-function visitAllProperties(data, modelCtorOrDef, visitor) {
+function visitAllProperties(data, modelCtor, visitor) {
   if (data === null || data === undefined) return;
-  const modelProps = modelCtorOrDef.properties ? modelCtorOrDef.properties : modelCtorOrDef.definition.properties;
+  const modelProps = modelCtor.properties ? modelCtor.properties : modelCtor.definition.properties;
   const allProps = new Set(Object.keys(data).concat(Object.keys(modelProps)));
   for (const p of allProps) {
     const value = data[p];
@@ -2024,28 +2068,47 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
         visitAllProperties(value, def.type.definition, visitor);
       }
     } else {
-      visitor(value, def, (newValue) => { data[p] = newValue; });
+      visitor(modelCtor, value, def, (newValue) => { data[p] = newValue; });
     }
     continue;
   }
 }
 
 /**
- *
- * @param {*} propValue Property value to coerce into a Decimal128 value
- * @param {*} propDef Property definition to check if property is MongoDB
- * Decimal128 type
+ * @param {*} modelCtor Model constructor
+ * @param {*} propValue Property value to coerce into special types supported by the connector
+ * @param {*} propDef Property definition to check if property is for MongoDB
  */
-function coerceDecimalProperty(propValue, propDef, setValue) {
-  let updatedValue;
-  if (hasDataType('decimal128', propDef)) {
-    if (Array.isArray(propValue)) {
-      updatedValue = propValue.map(val => Decimal128.fromString(val));
-      return setValue(updatedValue);
-    } else {
-      updatedValue = Decimal128.fromString(propValue);
-      return setValue(updatedValue);
+function coercePropertyValue(modelCtor, propValue, propDef, setValue) {
+  let coercedValue;
+  // Process only mongo-specific data types
+  if (propDef && propDef.mongodb && propDef.mongodb.dataType) {
+    const dataType = propDef.mongodb.dataType;
+    if (typeof dataType === 'string') {
+      if (hasDataType('decimal128', propDef)) {
+        if (Array.isArray(propValue)) {
+          coercedValue = propValue.map(val => Decimal128.fromString(val));
+          return setValue(coercedValue);
+        } else {
+          coercedValue = Decimal128.fromString(propValue);
+          return setValue(coercedValue);
+        }
+      } else if (typeIsObjectId(dataType)) {
+        if (isObjectIDProperty(modelCtor, propDef, propValue)) {
+          coercedValue = ObjectID(propValue);
+          return setValue(coercedValue);
+        } else {
+          throw new Error(`${propValue} is not an ObjectID string`);
+        }
+      }
+    } else if (dataType instanceof ObjectID) {
+      coercedValue = ObjectID(propValue);
+      return setValue(coercedValue);
     }
+  } else {
+    // Object ID coercibility depends on multiple factors, let coerceToObjectId() handle it
+    propValue = coerceToObjectId(modelCtor, propDef, propValue);
+    setValue(propValue);
   }
 }
 
@@ -2072,4 +2135,3 @@ function hasDataType(dataType, propertyDef) {
     propertyDef.mongodb.dataType &&
     propertyDef.mongodb.dataType.toLowerCase() === dataType.toLowerCase();
 }
-
--- a/test/id.test.js
+++ b/test/id.test.js
@@ -237,7 +237,6 @@ describe('mongodb default id name', function() {
         age: 30,
       },
       function(err, customer) {
-        // console.log(customer);
         customer.should.have.property('id');
         Customer1.findById(customer.id, function(err, customer1) {
           customer1.id.should.eql(customer.id);
@@ -245,5 +244,167 @@ describe('mongodb default id name', function() {
         });
       }
     );
+  });
+});
+
+describe('strictObjectIDCoercion', function() {
+  const ObjectID = ds.connector.getDefaultIdType();
+  const objectIdLikeString = '7cd2ad46ffc580ba45d3cb1f';
+
+  context('set to false (default)', function() {
+    const User = ds.createModel(
+      'user1',
+      {
+        id: {type: String, id: true},
+        name: String,
+      }
+    );
+
+    beforeEach(function(done) {
+      User.deleteAll(done);
+    });
+
+    it('should coerce to ObjectID', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+    });
+
+    it('should find model with ObjectID id', async function() {
+      const user = await User.create({name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+      const found = await User.findById(user.id);
+      found.toObject().name.should.equal('John');
+    });
+
+    it('should find model with ObjectID-like id', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+      user.id.should.eql(ObjectID(objectIdLikeString));
+      const found = await User.findById(objectIdLikeString);
+      found.toObject().name.should.equal('John');
+    });
+
+    it('should find model with string id', async function() {
+      const user = await User.create({id: 'a', name: 'John'});
+      user.id.should.be.an.instanceOf(String);
+      user.id.should.equal('a');
+      const found = await User.findById('a');
+      found.toObject().name.should.equal('John');
+    });
+  });
+
+  context('set to true', function() {
+    const User = ds.createModel(
+      'user2',
+      {
+        id: {type: String, id: true},
+        name: String,
+      },
+      {strictObjectIDCoercion: true}
+    );
+
+    beforeEach(function(done) {
+      User.deleteAll(done);
+    });
+
+    it('should not coerce to ObjectID', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.equal(objectIdLikeString);
+    });
+
+    it('should not find model with ObjectID id', async function() {
+      const user = await User.create({name: 'John'});
+      const found = await User.findById(user.id);
+      (found === null).should.be.true();
+    });
+
+    it('should find model with ObjectID-like string id', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.not.be.an.instanceOf(ds.ObjectID);
+      user.id.should.eql(objectIdLikeString);
+      const found = await User.findById(objectIdLikeString);
+      found.toObject().name.should.equal('John');
+    });
+
+    it('should find model with string id', async function() {
+      const user = await User.create({id: 'a', name: 'John'});
+      user.id.should.be.an.instanceOf(String);
+      user.id.should.equal('a');
+      const found = await User.findById('a');
+      found.toObject().name.should.equal('John');
+    });
+  });
+
+  context('set to true, id type set to ObjectID', function() {
+    const User = ds.createModel(
+      'user3',
+      {
+        id: {type: String, id: true, mongodb: {dataType: 'ObjectID'}},
+        name: String,
+      },
+      {strictObjectIDCoercion: true}
+    );
+
+    const User1 = ds.createModel(
+      'user4',
+      {
+        id: {type: String, id: true, mongodb: {dataType: 'objectid'}},
+        name: String,
+      },
+      {strictObjectIDCoercion: true}
+    );
+
+    beforeEach(function(done) {
+      User.deleteAll();
+      User1.deleteAll(done);
+    });
+
+    it('should coerce to ObjectID', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+    });
+
+    it('should coerce to ObjectID (lowercase)', async function() {
+      const user = await User1.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+    });
+
+    it('should throw if id is not a ObjectID-like string', async function() {
+      try {
+        await User.create({id: 'abc', name: 'John'});
+      } catch (e) {
+        e.message.should.match(/not an ObjectID string/);
+      }
+    });
+
+    it('should find model with ObjectID id', async function() {
+      const user = await User.create({name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+      const found = await User.findById(user.id);
+      found.toObject().name.should.equal('John');
+    });
+
+    // This works by coercing string to ObjectID, overriding `strictObjectIDCoercion: true`
+    it('should find model with ObjectID-like id', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+      user.id.should.eql(ObjectID(objectIdLikeString));
+      const found = await User.findById(objectIdLikeString);
+      found.toObject().name.should.equal('John');
+    });
+
+    it('should update model with ObjectID id', async function() {
+      const user = await User.create({name: 'John'});
+      await User.update({id: user.id, name: 'Jon'});
+      const found = await User.findById(user.id);
+      found.name.should.equal('Jon');
+    });
+
+    it('should update model with ObjectID-like id', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      await User.update({id: objectIdLikeString, name: 'Jon'});
+      const found = await User.findById(objectIdLikeString);
+      found.name.should.equal('Jon');
+    });
   });
 });
--- a/test/objectid.test.js
+++ b/test/objectid.test.js
@@ -7,11 +7,13 @@
 
 require('./init.js');
 
-let ds, Book, Chapter;
+let Book, Chapter;
+const ds = global.getDataSource();
+const ObjectID = ds.connector.getDefaultIdType();
+const objectIDLikeString = '7cd2ad46ffc580ba45d3cb1f';
 
 describe('ObjectID', function() {
   before(function() {
-    ds = global.getDataSource();
     Book = ds.define('Book');
     Chapter = ds.define('Chapter');
     Book.hasMany('chapters');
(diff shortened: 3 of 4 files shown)


test/id.test.js
```diff
@@ -247,3 +247,64 @@ describe('mongodb default id name', function() {
...
+
+    beforeEach(function(done) {
+      User.deleteAll(done);
+    });
+
+    it('should find model with ObjectID type id', function(done) {
+
+      User.create({ name: 'John' }, function(err, customer) {
+          User.find({where: {seq: {inq: [customer.id]}}}, function(err, customers) {
+            customers.should.not.be.empty();
```
hacksparrow: @bajtos I am trying to test the current behavior, but something seem amiss. Am I doing something wrong here and expecting something else?

bajtos: > `{where: {seq: {inq: [customer.id]}}}`

Where is `seq` property coming from? Did you mean `id` instead?

lib/mongodb.js
```diff
@@ -1881,7 +1881,11 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value, options) {
   ) {
     return true;
   } else if ('string' === typeof value) {
+    var properties = this._models[model] && this._models[model].properties;
+    if (properties.id && properties.id.isObjectID) return true;
```
bajtos: This condition does not look correct to me. You should be checking whether `prop.mongodb.dataType` is set to `ObjectID` (case-insensitively). Possibly also check if `prop.type` is `string`.

Note that we need to perform the coercion for all properties, not only primary keys. Most notably for foreign keys used by models with a belongsTo relation.

hacksparrow: @bajtos I have updated the implementation.

When `mongodb: {dataType: 'ObjectID'}` is set, any ObjectID-like id will be stored as an ObjectID, overriding the model-level setting of `strictObjectIDCoercion: true`. The id will be stored as an ObjectID, and it will always be resolved as an ObjectID. Therefore there is no problem when it comes to relations.

Regarding your other point "perform the coercion for all properties". Do we want to do that in this PR?

bajtos: >  The id is stored as an ObjectID, and it will always be resolved as an ObjectID. Therefore there is no problem when it comes to relations.

Could you please verify that claim for the following model definition?

```
const properties = {
    // id is explicitly typed as string
    id: {
     type: string,
     id: true,
    },

    // additional properties, not important
    name: 'string',
}

const settings: {
  // please test both options: idInjection disabled and enabled
  idInjection: false
}

const MyModel = ds.define('MyModel', properties, settings);
```

> Regarding your other point "perform the coercion for all properties". Do we want to do that in this PR?

Feel free to leave it out. I think that your changes in `MongoDB.prototype.isObjectIDProperty` should be enough to fix coercion for all properties, so it should be enough to add few tests to verify. Anyhow, let's focus on fixing primary keys first.

hacksparrow: Without specifying `mongodb: {dataType: 'ObjectID'}`?

>  I think that your changes in MongoDB.prototype.isObjectIDProperty should be enough to fix coercion for all properties, so it should be enough to add few tests to verify

That method is called only for ids at the moment. If we want to coerce all the properties, we'll have to refactor the code to loop through the properties, in many places. That's why I was wondering if we should do it in this PR.

bajtos: > Without specifying ``mongodb: {dataType: 'ObjectID'}`?

Sorry, I forgot to include that flag.

I think the following definition should be correct:

```js
const properties = {
    // id is explicitly typed as string
    id: {
     type: string,
     id: true,
     mongodb: {dataType: 'ObjectID'}
    },

    // additional properties, not important
    name: 'string',
}
```

I see you already have some tests in place for that, so maybe there are not that many new tests needed.

> > I think that your changes in MongoDB.prototype.isObjectIDProperty should be enough to fix coercion for all properties, so it should be enough to add few tests to verify
>
> That method is called only for ids at the moment. If we want to coerce all the properties, we'll have to refactor the code to loop through the properties, in many places. That's why I was wondering if we should do it in this PR.

Thanks for the explanation. I was naively assuming that `isObjectIDProperty` would be called for all properties.

I agree with you to leave support for ObjectID coercion in non-id properties out of scope of this pull request.

lib/mongodb.js
```diff
@@ -1881,14 +1881,25 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value, options) {
   ) {
     return true;
   } else if ('string' === typeof value) {
+    // values doesn't look like an ObjectID
+    if (!value.match(/^[0-9a-fA-F]{24}$/)) return false;
```
bajtos: What happens when the property is configured with `mongodb: {dataType: 'ObjectID'}` byt the value doesn't look like an ObjectID? Will we store it as a string in MongoDB? IMO, such behavior is not correct, because the developer explicitly asked the values to be stored as ObjectIDs. Operations trying to set a non-ObjectID values to ObjectID properties should fail, or at least behave the same way as when setting non-ObjectID values to properties configured as `{type: ObjectID}`. Please add tests to verify.

Notice that when `prop.type === ObjectID` (lines 1880-1882 above), the value is not checked and the property is always considered as having ObjectID. IMO, we should use the same logic for `{type: ObjectID}`,

lib/mongodb.js
```diff
@@ -1881,14 +1881,25 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value, options) {
...
+
+    // value looks like an ObjectID, and it should be treated as one
+    // because `prop.mongodb.dataType` is set to 'ObjectID'
+    // this condition overrides the `strictObjectIDCoercion` setting
+    if (prop.id && prop.mongodb && prop.mongodb.dataType === 'ObjectID') return true;
```
bajtos: Why is it important to check if `prop.id` is set? When `prop.type` is set to `ObjectID`, we consider the property as ObjectID-type regardless of `prop.id` value.

Can we do the same for `{mongodb: {dataType: 'ObjectID'}}`?

---

Also, should we support `dataType` set to ObjectID class constructor (as opposed to string name)? Should we support different casing of string values, e.g. `objectId`, `objectid`, etc.?

test/id.test.js
```diff
@@ -247,3 +247,182 @@ describe('mongodb default id name', function() {
     );
   });
 });
+
+describe('strictObjectIDCoercion', function() {
```
bajtos: We already have tests for strictObjectIDCoercion in [test/objectid.test.js](https://github.com/strongloop/loopback-connector-mongodb/blob/master/test/objectid.test.js).

Have you considered placing your new test to that file instead? If you prefer to keep them in `test/id.test.js`, then what rule should we use to decide where to put new ObjectID-related tests in the future (`test/id.test.js` vs. `test/objectid.test.js`)?

Equally important, please check the existing tests in `test/objectid.test.js` file and make sure your new tests are not duplicating any existing tests.

test/id.test.js
```diff
@@ -247,3 +247,182 @@ describe('mongodb default id name', function() {
...
+        id: {type: String, id: true},
+        name: String,
+      }
+    );
+
+    beforeEach(function(done) {
+      User.deleteAll(done);
+    });
+
+    it('should find model with ObjectID id', function(done) {
```
bajtos: Let's use `async` functions instead of callbacks in new tests please.

```js
it('should find model with ObjectID id', async () => {
  const user = await User.create({name: 'John'});
  user.id.should.be.an.instanceOf(ds.ObjectID);
  const found = await User.findById(user.id);
  user.should.not.be.null();
});
```

lib/mongodb.js
bajtos: On the second thought, I think the connector should always honor `mongodb: {dataType: 'ObjectID'}`, regardless of the actual value being set. I.e. the dataType check should be performed even for values that are not strings.

test/id.test.js
hacksparrow: Initially started with `test/objectid.test.js`, but moved to `test/id.test.js` because the tests were id-specific rather than ObjectID generic. May move it back eventually, let me focus on the implementation for now.

lib/mongodb.js
hacksparrow: > Why is it important to check if prop.id is set?

Oh, something from the previous implementation.

hacksparrow: @bajtos 

1. Currently `isObjectIDProperty` is called only for id properties. Should we apply it to all the properties? 
2. Currently `strictObjectIDCoercion` is applicable only to id properties. Should it be applicable to all the properties? 
3. There are no tests asserting the behavior of properties that are defined as ObjectID, but are assigned mismatching value. The current behavior is to try to convert to the type and store the result. Eg: for model property `xid: { type: 'number' }`, `xid: 'a'` will be stored as `'xid': NaN`. How do we respond to mismatched types? Throw?

I am in favor of supporting `dataType` set to `ObjectID` class constructor (as opposed to string name). This lays the foundation of supporting `ObjectID` at the higher levels eventually.

I think all of the above are desirable refactoring we should do, although it increase the scope of this PR.

lib/mongodb.js
```diff
@@ -2023,28 +2033,49 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
...
-function coerceDecimalProperty(propValue, propDef, setValue) {
-  let updatedValue;
-  if (hasDataType('decimal128', propDef)) {
-    if (Array.isArray(propValue)) {
-      updatedValue = propValue.map(val => Decimal128.fromString(val));
-      return setValue(updatedValue);
-    } else {
-      updatedValue = Decimal128.fromString(propValue);
-      return setValue(updatedValue);
+function coerceProperties(model, propValue, propDef, setValue) {
```
b-admike: I'd prefer having two distinct functions `coerceDecimalProperty` and `coerceObjectIdProperty`, but I understand that could be costly to call `visitAllProperties` twice.

lib/mongodb.js
```diff
@@ -2023,28 +2033,49 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
...
+        } else {
+          coercedValue = Decimal128.fromString(propValue);
+          return setValue(coercedValue);
+        }
+      }
+    // ObjectID data type
+    } else if (dataType === ObjectID) {
+      if (!model.settings.strictObjectIDCoercion) {
+        coercedValue = ObjectID(propValue);
+        setValue(coercedValue);
```
b-admike: `return setValue(coercedValue);` here?

lib/mongodb.js
```diff
@@ -2023,28 +2033,49 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
...
+      }
     }
+  } else {
+    // Apply strictObjectIDCoercion setting to ObjectID-like string values
+    if (typeof propValue === 'string' && !model.settings.strictObjectIDCoercion && propValue.match(/^[0-9a-fA-F]{24}$/)) {
+      coercedValue = ObjectID(propValue);
+      return setValue(coercedValue);
+    }
+
+    return setValue(propValue);
```
b-admike: Not sure why we need this return statement here.

lib/mongodb.js
b-admike: > On the second thought, I think the connector should always honor `mongodb: {dataType: 'ObjectID'}`, regardless of the actual value being set. I.e. the dataType check should be performed even for values that are not strings.

If this means we _try_ to coerce them to ObjectID, then I am okay with it. If it's not an ObjectID like string, then coercion should throw an error, right?

bajtos: > If it's not an ObjectID like string, then coercion should throw an error, right?

Yes, that's what I would like to see. Here is my thinking:

- The application developer configured the model property to be stored as ObjectID in MongoDB.
- The client provided property value that's not a valid ObjectID.
- My opinion: the framework should reject the request as invalid.

IMO, when a developer configures as property as ObjectID, then the framework should ensure that only ObjectID value are ever stored in that property. That way the application developer can rely on the assumption that the string values will be always in ObjectID format and thus can keep their code simple (no need to check everywhere whether the string is really an ObjectID).

lib/mongodb.js
```diff
@@ -1881,14 +1881,30 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value, options) {
...
     return true;
   } else if ('string' === typeof value) {
+    // dataType is specified as ObjectID
+    if (prop && prop.mongodb && prop.mongodb.dataType === ObjectID) {
+      // value looks like an ObjectID, and it should be treated as one
+      // because `prop.mongodb.dataType` is set to 'ObjectID'
+      // this condition overrides the `strictObjectIDCoercion` setting
+      if (value.match(/^[0-9a-fA-F]{24}$/)) return true;
+      // if dataType is specified as ObjectID, value is expected to an ObjectID-like string
+      else throw new Error(`${value} is not an ObjectID string`);
```
bajtos: IMO, this error should be thrown by the code performing coercion from string to ObjectID instance, not by the helper checking whether a property expects ObjectID values.

`isObjectIDProperty` should simply return "true" when the property value should be stored as ObjectID and "false" when not, regardless of whether the actual value can be converted to ObjectID or not.

For example, clients can provide a numeric value for an ObjectID property. Such request will not go through this line, and yet we want to reject it as invalid. BTW, please make sure there is a test case for this scenario.

lib/mongodb.js
```diff
@@ -1881,14 +1881,30 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value, options) {
...
+
     options = options || {};
     var strict =
       (settings && settings.strictObjectIDCoercion) ||
       this.settings.strictObjectIDCoercion ||
       options.strictObjectIDCoercion;
+
     if (strict) return false; // unless explicitly typed, don't coerce
-    return /^[0-9a-fA-F]{24}$/.test(value);
+    else return true;
```
bajtos:
```diff
- if (strict) return false; // unless explicitly typed, don't coerce
- else return true;
+ // coerce only when strictObjectIDCoercion is disabled
+ return !strict;
```

lib/mongodb.js
```diff
@@ -1881,14 +1881,30 @@ MongoDB.prototype.isObjectIDProperty = function(model, prop, value, options) {
   ) {
     return true;
   } else if ('string' === typeof value) {
+    // dataType is specified as ObjectID
+    if (prop && prop.mongodb && prop.mongodb.dataType === ObjectID) {
+      // value looks like an ObjectID, and it should be treated as one
+      // because `prop.mongodb.dataType` is set to 'ObjectID'
+      // this condition overrides the `strictObjectIDCoercion` setting
+      if (value.match(/^[0-9a-fA-F]{24}$/)) return true;
```
bajtos: This regular expression is repeated in quite few places, please extract it into a shared constant.

lib/mongodb.js
```diff
@@ -2000,15 +2016,14 @@ function optimizedFindOrCreate(model, filter, data, options, callback) {
...
 /**
- *
  * @param {*} data Plain Data Object for the matching property definition(s)
- * @param {*} modelCtorOrDef Model constructor or definition
+ * @param {*} model Model of the data
  * @param {*} visitor A callback function which takes a property value and
  * definition to apply custom property coercion
  */
-function visitAllProperties(data, modelCtorOrDef, visitor) {
+function visitAllProperties(data, model, visitor) {
```
bajtos: The name `model` is ambiguous, it can mean "model constructor" but also "model instance". Please use `modelCtor` instead.

Same comment applies to other places where you are introducing `model` name.

lib/mongodb.js
```diff
@@ -2023,28 +2033,49 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
...
-    } else {
-      updatedValue = Decimal128.fromString(propValue);
-      return setValue(updatedValue);
+function coerceProperties(model, propValue, propDef, setValue) {
+  let coercedValue;
+  // Process only mongo-specific data types
+  if (propDef && propDef.mongodb && propDef.mongodb.dataType) {
+    const dataType = propDef.mongodb.dataType;
+    // Data types specified as string values
+    // Yaapa: should we implement all type specifications using objects instead of strings?
```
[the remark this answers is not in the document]
bajtos: I am not sure what's the actual question here.

Are you asking about MongoDB-specific types other than `ObjectID` and `Decimal128`, e.g. `timestamp` (see [MongoDB's doc on BSON types](https://docs.mongodb.com/manual/reference/bson-types/index.html))? IMO, it would be great eventually support all BSON types, but such changes are definitely out of scope of this pull request.

lib/mongodb.js
```diff
@@ -2023,28 +2038,49 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
...
+    // ObjectID data type
+    } else if (dataType === ObjectID) {
+      if (!model.settings.strictObjectIDCoercion) {
+        coercedValue = ObjectID(propValue);
+        setValue(coercedValue);
+      }
     }
+  } else {
+    // Apply strictObjectIDCoercion setting to ObjectID-like string values
+    if (typeof propValue === 'string' && !model.settings.strictObjectIDCoercion && propValue.match(/^[0-9a-fA-F]{24}$/)) {
```
bajtos: Why are we checking `model.settings.strictObjectIDCoercion` again? The rule to decide whether a string value should be coerced to ObjectID is more complex than a RegExp check, that's why we have `isObjectIDProperty`.

hacksparrow:
I am fine with the increased scope so far, the features are not too different from each other. Agreed, we should take this opportunity to refactor the codebase.

test/objectid.test.js
```diff
@@ -32,41 +34,75 @@ describe('ObjectID', function() {
...
+      it('if type is set to ObjectID, should coerce to ObjectID', async function() {
+        Book = ds.createModel(
+          'book',
+          {
+            xid: {type: String, mongodb: {dataType: ObjectID}}
+          },
+          {strictObjectIDCoercion: true}
+        );
+        const book = await Book.create({xid: objectIDLikeString});
+        book.xid.should.be.an.instanceOf(ds.ObjectID);
```
hacksparrow: @b-admike any idea why this fails despite the line 2095 being executed in `/lib/mongodb.js`?

test/objectid.test.js
```diff
@@ -32,41 +34,75 @@ describe('ObjectID', function() {
...
+
+      it('it should coerce to ObjectID', async function() {
+        Book = ds.createModel(
+          'book',
+          {
+            xid: {type: String}
+          }
+        );
+        const book = await Book.create({xid: objectIDLikeString});
+        book.xid.should.be.an.instanceOf(ds.ObjectID);
```
hacksparrow: @b-admike any idea why this fails despite the line 2112 being executed in `/lib/mongodb.js`?

jannyHou:
> So perhaps we should use this opportunity to clean up and refactor the coercion code? Rework as many places as possible to leverage visitAllProperties helper, and get rid of any hacky code that we accumulated in the past.

I would respect Yaapa's preference since he is the PR owner 🙂speak for myself, +1 for cleaning up the coercion code.

test/objectid.test.js
b-admike: Yea Janny's right. I had to query the db myself and get the raw results back (see https://github.com/strongloop/loopback-connector-mongodb/blob/master/test/decimal.test.js#L468-L470) to make the assertions.

hacksparrow: I see. So we have a sort of confusing behavior, here. In fact, I remember now, I was thinking of coercing in the properties the result too.

Now, the question is. Should the query result preserve the ObjectID type? In my opinion, we should.

@jannyHou @b-admike @bajtos @raymondfeng

b-admike:
> Besides my comments below, I am concerned about the growing scope of this pull request. At the moment, I see two changes:
> 
> * Improve condition deciding whether a string value should be coerced to ObjectID, recognize `{mongodb: {dataType: 'ObjectID'}}`.
> * Refactor coercion of (non-id) property values to apply ObjectID coercion too.
> 
> I am not sure if it's a good idea to keep them in a single pull request?
> 
> On the other hand: I believe the currently proposed implementation of coercion of non-id property values could break existing applications. I think that's fine, because IMO the proposed behavior is more consistent and less surprising, but it also means we need to release these changes as a semver-major release.
> 
> So perhaps we should use this opportunity to clean up and refactor the coercion code? Rework as many places as possible to leverage `visitAllProperties` helper, and get rid of any hacky code that we accumulated in the past.
> 
> @hacksparrow @jannyHou @b-admike WDYT?

Yea I would prefer to keep the scope of the PR as small as possible (or at least have multiple logical commits). I'm not sure if coercing non-id ObjectID properties is something that aligns with MongoDB principles like @jannyHou mentioned. If we are good in that front, then the proposal looks good to me. +1 on releasing a semver-major release if that's the case. I think we should identify areas that need cleanup from this point forward and follow-up with stories that we can tackle after. The changes so far are reasonable and have already made significant changes/improvements. If what's left won't require lots of effort, then yes it could be beneficial to wrap it up and release it with the semver-major changes. Let's also update the README to clearly state what changes to our users since we have a section on `strictObjectIDCoercion` flag there.

lib/mongodb.js
```diff
@@ -180,6 +184,8 @@ function MongoDB(settings, dataSource) {
   }
 
   this.dataSource = dataSource;
+  connectorSettings = this.settings;
```
bajtos: This isn't going to work. There can be multiple connector instances used by a single application (a single Node.js process).

Please find a different way how to access connector settings from functions that are not bound to the prototype.

lib/mongodb.js
```diff
@@ -336,41 +342,41 @@ MongoDB.prototype.getDefaultIdType = function() {
 
 /**
  * Get collection name for a given model
- * @param {String} model Model name
+ * @param {String} modelName Model name
  * @returns {String} collection name
  */
-MongoDB.prototype.collectionName = function(model) {
-  var modelClass = this._models[model];
+MongoDB.prototype.collectionName = function(modelName) {
```
bajtos: Please extract renames into a standalone pull request. When renames are mixed with code changes that are changing behavior, it's easy to miss an important change during the review. Let's quickly land the renames first and them make this pull request much smaller in patch size.

lib/mongodb.js
```diff
@@ -400,30 +406,30 @@ MongoDB.prototype.fromDatabase = function(model, data) {
...
     // Override custom column names
-    data = this.fromPropertyToDatabaseNames(model, data);
+    data = this.fromPropertyToDatabaseNames(modelName, data);
     return data;
   }
 
-  for (var p in props) {
-    var prop = props[p];
+  for (const p in props) {
+    const prop = props[p];
```
bajtos: While you are changing from `var` to `const`, I think it would be best to open a new pull request to update `eslint-config-loopback` to the latest version and let `eslint` to fix all `var` usages. See https://github.com/strongloop/eslint-config-loopback/releases/tag/v13.0.0

lib/mongodb.js
```diff
@@ -537,7 +544,18 @@ MongoDB.prototype.coerceId = function(model, id, options) {
...
+      if (isObjectIDProperty(modelCtor, idProp, idValue, options)) {
+        idValue = ObjectID(idValue);
+      } else {
+        throw new Error(`${idValue} is not an ObjectID string`);
+      }
+    } else if (enableStrictObjectIDCoercion(modelCtor, options)) {
+      if (isObjectIDProperty(modelCtor, idProp, idValue, options)) {
+        idValue = ObjectID(idValue);
+      }
+    } else if (isObjectIDProperty(modelCtor, idProp, idValue, options)) {
```
bajtos: This code looks much better than before.

It makes me wonder though - do we still need this special rule in `coerceId` function? Would the coercion be already handled by `visitAllProperties`?

If there are cases that would not be handled by `visitAllProperties`, then please ensure there are tests covering them.

lib/mongodb.js
```diff
@@ -1869,26 +1886,49 @@ MongoDB.prototype.ping = function(cb) {
   }
 };
 
+// Determine if a property must have an ObjectID value
+function shouldBeObjectID(propDef) {
+  if (!propDef) return false;
+  if (propDef.type === ObjectID || propDef.mongodb && propDef.mongodb.dataType === ObjectID) return true;
```
bajtos: I believe that `dataType` is usually going to be a string value (`{dataType: 'ObjectID'}`), not a class constructor. Unless you have code that's normalizing the dataType value from a string to a constructor? 

Please add tests to verify.

lib/mongodb.js
```diff
@@ -1869,26 +1886,49 @@ MongoDB.prototype.ping = function(cb) {
...
-    var settings = this._models[model] && this._models[model].settings;
-    options = options || {};
-    var strict =
-      (settings && settings.strictObjectIDCoercion) ||
-      this.settings.strictObjectIDCoercion ||
-      options.strictObjectIDCoercion;
-    if (strict) return false; // unless explicitly typed, don't coerce
-    return /^[0-9a-fA-F]{24}$/.test(value);
+    // dataType is specified as ObjectID
+    if (propDef && propDef.mongodb && propDef.mongodb.dataType === ObjectID) {
```
bajtos: Ditto, I believe that `dataType` will be usually a string `ObjectID`.

lib/mongodb.js
```diff
@@ -2023,28 +2062,57 @@ function visitAllProperties(data, modelCtorOrDef, visitor) {
...
-function coerceDecimalProperty(propValue, propDef, setValue) {
-  let updatedValue;
-  if (hasDataType('decimal128', propDef)) {
-    if (Array.isArray(propValue)) {
-      updatedValue = propValue.map(val => Decimal128.fromString(val));
-      return setValue(updatedValue);
-    } else {
-      updatedValue = Decimal128.fromString(propValue);
-      return setValue(updatedValue);
+function coerceProperties(modelCtor, propValue, propDef, setValue) {
```
bajtos: I find the name misleading. It says "coerce properties" (plural name), but it's coercing a single property value AFAICT.

Proposed name:
 
```suggestion
function coercePropertyValue(modelCtor, propValue, propDef, setValue) {
```

bajtos:
> The id coercion related logic in this PR makes sense to me(correct me if my understanding is wrong):
> 
> When store data into the database:
> - if  `{type: ObjectId}`: coerce from string to ObjectId.
> - If `{mongodb: {dataType: ObjectId}}` and data looks like objectId, coerce from string to ObjectId.
> - If no property configurations set, data looks like objectId, and `strictObjectIDCoercion` disabled, coerce from string to ObjectId.

Great summary. Personally, I am arguing for a slightly different behavior though:

- If `{mongodb: {dataType: ObjectId}}`, **always** coerce from string to ObjectId.

> The type of returned data from juggler only depends on the model property's type:
> - If `{type: String}` return a string property
> - If `{type: ObjectId}` return an ObjectId property

👍 

> > Refactor coercion of (non-id) property values to apply ObjectID coercion too
>
> Interesting...I am not sure why a property need to be stored in type ObjectId when it is NOT an id property...in mongodb, type objectId already implies the field is an id field IMO...

Consider HasMany/BelongsTo relation "Category has many Product instances", where Product contains `categoryId` property. 

If both models are backed by MongoDB, I think we need to ensure that `categoryId` property is always set to `ObjectID` value, even if the caller provided a string only (e.g. because `categoryId` was obtained from HTTP request path like `/api/categories/{some-objectid-string}/products`. If we don't coerce, then we may end up with running a query that's providing ObjectID instance in the condition for `categoryId`, but the model uses a string value for `categoryId`. My assumption is that such query will not find any records. Of course, if MongoDB is clever enough to coerce the values during comparison, then we may not need to coerce them ourselves.

test/objectid.test.js
bajtos: I don't have a strong opinion. I agree the current behavior is confusing. To me, it's important to leave any fixes out of scope of this pull request. Considering how much time we have already spend on fixing the MongoDB ObjectID coercion issue, I am reluctant to get distracted by other improvements. Let's open a new GitHub issue to describe the problem so that we can properly plan the time needed to fix it; and rework the tests to accommodate limitations of the current implementation.

lib/mongodb.js
```diff
@@ -942,12 +946,12 @@ MongoDB.prototype.buildWhere = function(modelName, where, options) {
...
-    if (isDecimal) {
-      cond = Decimal128.fromString(cond);
-      debug('buildWhere decimal value: %s, constructor name: %s', cond, cond.constructor.name);
+    if (prop && prop.mongodb && typeof prop.mongodb.dataType === 'string') {
+      const isDecimal = prop.mongodb.dataType.toLowerCase() === 'decimal128';
+      if (isDecimal) {
+        cond = Decimal128.fromString(cond);
+        debug('buildWhere decimal value: %s, constructor name: %s', cond, cond.constructor.name);
+      }
     }
```
bajtos: I think you need to handle `prop.mongodb.dataType.toLowerCase() === 'objectid'` too, so that string properties stored as ObjectIDs are handled in `where` queries too. Please start with a unit test that fails now.

I am ok to leave this part out of scope of this pull request if you prefer, but I think it should be implemented before we can call the user story done.

lib/mongodb.js
```diff
@@ -1870,30 +1873,71 @@ MongoDB.prototype.ping = function(cb) {
   }
 };
 
+// Determine if a property must have an ObjectID value
+function mustBeObjectID(propDef) {
```
bajtos: Why `must`? The name implies to me that an error is thrown when `propDef` is not ObjectID`. Proposed name:

```suggestion
function isStoredAsObjectID(propDef) {
```

lib/mongodb.js
```diff
@@ -1870,30 +1873,71 @@ MongoDB.prototype.ping = function(cb) {
...
+  if (!propDef) return false;
+  if (propDef.mongodb && ObjectIdTypeRegex.test(propDef.mongodb.dataType)) return true;
+  return false;
+}
+
+// Determine if strictObjectIDCoercion should be enabled
+function enableStrictObjectIDCoercion(modelCtor, options) {
```
bajtos: The name implies the function is going to enable coercion, which is not true AFAICT. Proposed name:

```suggestion
function isStrictObjectIDCoercionEnabled(modelCtor, options) {
```

lib/mongodb.js
```diff
@@ -1870,30 +1873,71 @@ MongoDB.prototype.ping = function(cb) {
...
+  const settings = modelCtor.settings;
+  return (settings && settings.strictObjectIDCoercion) ||
+    (modelCtor.model && modelCtor.model.getConnector().settings.strictObjectIDCoercion) ||
```
bajtos: I find this unnecessarily complicated. Can you change `enableStrictObjectIDCoercion` into a connector method, so that you can access connector settings via `this.settings`? Obviously, other functions calling this one will have have to be converted into connector methods too.

lib/mongodb.js
```diff
@@ -1870,30 +1873,71 @@ MongoDB.prototype.ping = function(cb) {
...
+  }
+  return propValue;
+}
+
 /**
  * Check whether the property is an ObjectID (or Array thereof)
  *
  */
-MongoDB.prototype.isObjectIDProperty = function(modelName, prop, value, options) {
+function isObjectIDProperty(modelCtor, propDef, value, options) {
```
bajtos: What's the difference between `isObjectIDProperty` and `mustBeObjectID`? Can we merge those two methods into a single one?

jannyHou:
> If both models are backed by MongoDB, I think we need to ensure that categoryId property is always set to ObjectID value, even if the caller provided a string only (e.g. because categoryId was obtained from HTTP request path like /api/categories/{some-objectid-string}/products. 

@bajtos Thank you for the detailed explanation 👍 I forgot the relation constraint. Fair enough.

lib/mongodb.js
hacksparrow: You are correct, if `mustBeObjectID(propDef) === true`, an error will be thrown by `coerceToObjectId()` if the value is not an `ObjectID`.

`isStoredAsObjectID` does not convey the intended message, IMO.

lib/mongodb.js
hacksparrow: `mustBeObjectID` determines whether a property's value must be `ObjectID`.  It is like specifying the data type.

`isObjectIDProperty` checks if the property value is an `ObjectID`. It is like checking the data type. This is how it does the determination:

- If the value doesn't look like an ObjectID -> `false`
- If the value looks like an ObjectID and `strictObjectIDCoercion` is set to `false` (default) and `mongodb.dataType` is not set -> `true`
- If the value looks like an ObjectID and `strictObjectIDCoercion` is set to `true` and `mongodb.dataType` is not set -> `false`
- If the value looks like an ObjectID and `strictObjectIDCoercion` is set to `true` and `mongodb.dataType` is set to "ObjectID" -> true

I actually created these two functions out of a single function to add clarity to the flow. I am not in favor on re-merging them, since they serve different purposes and can be used independently in other places.

lib/mongodb.js
hacksparrow: I agree, it is not simple, but it is not too complicated at the same time.

The following helper functions have to be added to the prototype to make `this.settings` possible:

1. `coercePropertyValue`
2. `coerceToObjectId`
3. `isObjectIDProperty`
4. `isStrictObjectIDCoercionEnabled`

Does the ability to use `this.settings` instead of `modelCtor.model && modelCtor.model.getConnector().settings`, out weight the addition of those extra methods to the prototype?

I initially started with the approach you suggested but decided otherwise, because I found it very inelegant in this case.

lib/mongodb.js
hacksparrow: Given:

`const ObjectIdTypeRegex = /objectid/i;` 

Shouldn't `ObjectIdTypeRegex.test()` take care of things?

lib/mongodb.js
```diff
@@ -1870,30 +1873,71 @@ MongoDB.prototype.ping = function(cb) {
...
+      return ObjectID(propValue);
+    } else {
+      throw new Error(`${propValue} is not an ObjectID string`);
+    }
+  } else if (isStrictObjectIDCoercionEnabled(modelCtor)) {
+    if (isObjectIDProperty(modelCtor, propDef, propValue)) {
+      return ObjectID(propValue);
+    }
+  } else if (ObjectIdValueRegex.test(propValue)) {
+    return ObjectID(propValue);
```
jannyHou: Here we don't need to test `if (isObjectIDProperty(modelCtor, propDef, propValue))`?

hacksparrow: `ObjectIdValueRegex.test()` is checking if the string looks like an ObjectID.

In other places, we do `isObjectIDProperty()`, because whether a string is an ObjectID or not is determined by the following factors:

1. String pattern matches that of ObjectID
2. `strictObjectIDCoercion`
3. `dataType: "ObjectID"`

By line `1903`, factors 2 and 3 are not applicable, and the string looks like an ObjectID, so we simply coerce etc.

hacksparrow: @jannyHou is it clear to you now?

jannyHou: @hacksparrow Thanks for the answer fair enough 👍

test/objectid.test.js
b-admike: Agreed on deferring this to another issue. In terms of the behaviour, the query result is actually coerced for decimal types. It's just juggler which converts it back to string values. I think this is because, say, for explorer, we need to have it displayed as a string value.

test/objectid.test.js
```diff
@@ -55,18 +57,19 @@ describe('ObjectID', function() {
...
-    const coercedId = ds.connector.isObjectIDProperty(
-      'Book',
-      {},
-      '52fcef5c0325ace8dcb7a0bd',
-      {strictObjectIDCoercion: true}
-    );
-    coercedId.should.be.False();
+  context('strictObjectIDCoercion', function() {
+    context('set to true', function() {
+      it('should not coerce to ObjectID', async function() {
```
b-admike: I feel like the tests replaced by your new test are more like unit tests for `isObjectIDProperty`, and your new one is more like an acceptance test (I guess it does call `isObjectIDProperty` behind the scenes). Can we preserve the old tests?

test/id.test.js
```diff
@@ -247,3 +246,152 @@ describe('mongodb default id name', function() {
...
+    });
+
+    it('should coerce to ObjectID', async function() {
+      const user = await User.create({id: objectIdLikeString, name: 'John'});
+      user.id.should.be.an.instanceOf(ds.ObjectID);
+    });
+
+    it('should throw if id is not a ObjectID-like string', async function() {
+      // Why is this not working?
+      // await User.create({id: 'abc', name: 'John'}).should.be.rejectedWith(/not an ObjectID string/);
```
b-admike: Here are two ways of doing it (one is for expect/chai, one for mocha):
expect:
https://github.com/strongloop/loopback-next/blob/75939a4a144b3068ff2890394c178faad58d7458/packages/cli/test/integration/generators/relation.integration.js#L39-L48

Mocha:
https://github.com/strongloop/loopback-datasource-juggler/blob/91ab3e916281e56e8675889ab7c5a31dc223c22c/test/datasource.test.js#L446-L451

I suggest trying to have a promise based test:
```js
it('should throw if id is not an ObjectID-like string', () => {

  return user.create({id: 'abc', name: 'John'}).should.be.rejectedWith(/not an ObjectID string/);
});
```

lib/mongodb.js
b-admike: Well if the error is being thrown by `coerceToObjectId()`, I think it makes sense to rename this function to something like `isObjectIDType` (fine with `isStoredAsObjectID`) since objectively looking at this function alone, it's used to determine if the data type is `ObjectID`.

lib/mongodb.js
b-admike: We are actually checking the data type in `isObjectIDProperty` (see line 1922) so maybe we can call `mustBeObjectID` for it there. That function is called only once on line 1893 btw. My 2 cents is to keep both and perhaps rename `mustBeObjectID` to a more descriptive name.

test/objectid.test.js
hacksparrow: Previously `isObjectIDProperty` was a method on the prototype, now it is not. I removed it because I wanted to avoid adding helper functions to the prototype.

If we want to unit test the helper function `isObjectIDProperty`, shouldn't we test the rest of the helper functions too? We should ideally, but it seems like an overkill to me, because these helper functions are not exposed as connector methods and their functionality is getting tested via the integration tests, anyway. And to unit test these helper functions, we will have to add all of them to the prototype; I am not in favor of doing that.

@bajtos thoughts?

test/id.test.js
hacksparrow: > return user.create({id: 'abc', name: 'John'}).should.be.rejectedWith(/not an ObjectID string/);

Doesn't help.

Would it be ok to replace `should` with `expect` in this repo?

lib/mongodb.js
hacksparrow: > so maybe we can call mustBeObjectID for it there

Thanks, done.

lib/mongodb.js
hacksparrow: Renamed to `isStoredAsObjectID `.

test/objectid.test.js
b-admike: I see. Thanks for the background context on this @hacksparrow. I understand now it's not a prototype method anymore, but the tests _did_ exist for it, and I'm just asking to preserve them. We don't have to necessarily write unit tests for all the helper methods. I'm okay to not recover those tests too.

test/id.test.js
b-admike: Maybe my example is wrong. I didn't get to try it out. I'll try it out and let you know. In terms of replacing `should` with `expect`, it is ok, but not sure how much effort that is and how much beneficial it is.

test/objectid.test.js
hacksparrow: > I'm just asking to preserve them

@b-admike Preserve, as in "don't get rid of the old test"? If yes, what is the point of keeping it when we already know it will fail, since the test calls a now non-existent method?

I can understand your concern that a unit test is gone, but that particular condition and many more are now checked from a higher level.

The reason why I removed `isObjectIDProperty` from the prototype is because:

1. I wanted to modularize ObjectID-related functionality in the whole module, so had to created smaller helper functions, which are can be used individually too, and adds clarity to the flow.
2. If `isObjectIDProperty` was to be on the prototype, some of the helper functions had to be there too.
3. Removing `isObjectIDProperty` from the prototype was the more elegant approach than adding some more helper functions to the prototype, IMO.

lib/mongodb.js
bajtos: > Shouldn't `ObjectIdTypeRegex.test()` take care of things?

What will happen when the value is not an ObjectID string? IIUC, in your proposal the query will return no records, because the non-ObjectID string will not match any of the ObjectID values stored in the database.

Are we fine with that behavior?

I was thinking that we should reject such requests with `400 Bad Request` error instead.

Also, what happens when `strictObjectIDCoercion` is enabled and the property is defined as `{type: 'string', mongodb: {dataType: 'ObjectID'}}`. Will we correctly coerce the string value to ObjectID, so that it can match any existing values in the database?

lib/mongodb.js
bajtos: Fair enough, I can live with the current proposal.

lib/mongodb.js
bajtos: I find this problematic. IMO we should not make any distinction between properties defined as `{type: 'ObjectID'}` and `{type: 'string', mongodb: {dataType: 'ObjectID'}}`.

In both cases, the property should be considered by the connector as having ObjectID type.

My main concern is that by having two functions, it's easy to use a wrong one and introduce situation where certain code paths don't respect `mongodb: {dataType: 'ObjectID'}` setting.

lib/mongodb.js
hacksparrow: I got your point, I was thinking of something else.

lib/mongodb.js
hacksparrow: Agreed.

lib/mongodb.js
b-admike: > > Shouldn't `ObjectIdTypeRegex.test()` take care of things?
> 
> What will happen when the value is not an ObjectID string? IIUC, in your proposal the query will return no records, because the non-ObjectID string will not match any of the ObjectID values stored in the database.
> 
> Are we fine with that behavior?
> 
> I was thinking that we should reject such requests with `400 Bad Request` error instead.
> 
> Also, what happens when `strictObjectIDCoercion` is enabled and the property is defined as `{type: 'string', mongodb: {dataType: 'ObjectID'}}`. Will we correctly coerce the string value to ObjectID, so that it can match any existing values in the database?

I misunderstood this, sorry. Rejecting non-objectID string makes sense to me (I thought this was the behaviour). Also, for the case when strict coercion flag is enabled, we should coerce string values to ObjectID, and also reject/throw an error if we can't (invalid values).

lib/mongodb.js
b-admike: I get it now too. Yea we should aim at having unified logic for those code paths 👍 .

test/objectid.test.js
b-admike: Thank you for clarifying @hacksparrow. Ah I thought the tests could be recovered with the changes that took place. I'm good with leaving them out.

lib/mongodb.js
b-admike: @bajtos @hacksparrow is this something we can create a follow-up story on?

lib/mongodb.js
b-admike: @bajtos @hacksparrow can we create a follow-up story for this?

lib/mongodb.js
hacksparrow: Let's create a follow up story.

lib/mongodb.js
hacksparrow: I have fixed this one.

hacksparrow:
@b-admike any idea why this is failing? https://travis-ci.org/strongloop/loopback-connector-mongodb/jobs/540672910

hacksparrow:
Ok, all green now. Let's land this today.

mikeevstropov:
```
@property({
  mongodb: {dataType: 'ObjectID'},
  default: null,
  jsonSchema: {nullable: true},
})
```
```
Error: null is not an ObjectID string
```

lib/mongodb.js
taicho: @b-admike @hacksparrow @bajtos 
Created #545 as I believe it relates to this conversation. This commit and its related code appears to fix the issue related to string-properties-stored-as-ObjectId but does not account for a property that is an Array of ObjectID or Array of ObjectID-like-strings.

PR #519, opened 2019-05-13
Run shared tests from both v3 and v4 of juggler

bajtos:
### Description

At the moment, our connectors are installing juggler 3.x to run the test suite. When we make a change to juggler@3, cis-jenkins detects downstream dependency and triggers CI build of connectors with the modified juggler version. When we make a change to juggler's master, no downstream builds are triggered.

In the past, we were maintaining multiple branches to test connectors against different juggler versions. E.g. loopback-2.x containing code from "master" but installing juggler 2.x for testing. This is rather impractical, because we don't have bandwidth to update those other branches with changes made to master.

In this pull request, I am reworking connector test setup to execute tests from both juggler versions 3.x and 4.x.

#### Related issues

- Spike story: strongloop-internal/scrum-apex#416
- See also: https://github.com/strongloop/loopback-connector-postgresql/pull/376
- This PR requires https://github.com/strongloop/loopback-datasource-juggler/pull/1739 to be landed & released first

### Checklist


bajtos:
@slnode test please

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "benchmarks": "make benchmarks",
     "leak-detection": "make leak-detection",
-    "test": "mocha",
+    "test": "mocha test/*.test.js node_modules/juggler-v3/test.js node_modules/juggler-v4/test.js",
```
b-admike: Can you please explain why this works with `node_modules/juggler-v3/test.js`? I was expecting `deps/juggler-v3/test.js`.

bajtos: FWIW, on Node.js 8+, this works with `deps/juggler-v3/tests.js` too.

What happens under the hood: in `package.json`, we specify a dependency on `juggler-v3` as follows:

```
"juggler-v3": "file:./deps/juggler-v3"
```

When we run `npm install`, npm adds juggler-v3 to `node_modules` and installs its dependencies.

In npm shipped with Node.js 6, juggler-v3 would be copied to `node_modules/juggler-v3`. As a result, there will be no `deps/juggler-v3/node_modules` folder. Calling `require('loopback-datasource-juggler')` from `deps/juggler-v3/test.js` would resolve to `node_modules/loopback-datasource-juggler`, which is typically v4 - not what we want!

Newer npm versions create a symlink:

```
node_modules/juggler-v3 -> deps/juggler-v3
```

With the symlink in place, it does not really matter where we load `juggler-v3/test.js` from.

I prefer to keep loading from `node_modules` because I consider it as more correct conceptually.

PR #523, opened 2019-05-24
chore: replace var with let and const

hacksparrow:
### Description

Changes made as per [this discussion](https://github.com/strongloop/loopback-connector-mongodb/pull/517#discussion_r286900293).

`eslint --fix .` was ran on the repo. Whatever was not fixed by `eslint` was manually fixed. `npm test` was ran to ensure, nothing is broken as a result of these changes.

### DIFF
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -5,29 +5,29 @@
 
 'use strict';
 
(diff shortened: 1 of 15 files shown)


jannyHou:
LGTM 👍

PR #524, opened 2019-06-14
Show illegal unescaped password error on console

alejandrochvs:
When the password contains an illegal unescaped character there is no indication that the problem is that, this should indicate the problem and handle the exception

#### Related issues

### Checklist

- [ x] Code conforms with the [style

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -310,6 +310,9 @@ MongoDB.prototype.connect = function(callback) {
         self.client = client;
         // The database name might be in the url
         return urlParser(self.settings.url, self.settings, function(err, url) {
+          if (err) {
+            return console.error(err);
+          }
           self.db = client.db(
             url.dbName || self.settings.database,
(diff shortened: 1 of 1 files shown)


slnode:

hacksparrow:
We plan to handle the encoding - https://github.com/strongloop/loopback-connector-mongodb/pull/527

hacksparrow:
Thanks for the PR @alejandrochvs. Since we have closed https://github.com/strongloop/loopback-connector-mongodb/pull/527, let's consider this PR.

@bajtos @b-admike @jannyHou how do you suggest we handle this case?

In automated remote environments, `console.log()` may not help. Should the app continue to start with invalid creds or exit with a code?

hacksparrow:
If there is an encoding problem in the username or password, the whole error stack is printed on the console and the app will not start.

I don't think, we should do anything more. @alejandrochvs do let me know, if I misunderstood anything.

PR #525, opened 2019-06-14
fix: ObjectID data type preservation

hacksparrow:
Preserve ObjectID data type preservation across all properties

### Description

If a property is defined as `ObjectID` type, it is enforced on all the properties on a model.

#### Related issues

- https://github.com/strongloop/loopback-connector-mongodb/pull/517

### Checklist

### DIFF
--- a/test/objectid.test.js
+++ b/test/objectid.test.js
@@ -9,7 +9,6 @@ require('./init.js');
 
 let Book, Chapter;
 const ds = global.getDataSource();
-const ObjectID = ds.connector.getDefaultIdType();
 const objectIDLikeString = '7cd2ad46ffc580ba45d3cb1f';
 
 describe('ObjectID', function() {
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,4 @@
 node_modules
 coverage
 .idea
-
+.nyc_output/
```
bajtos: Please extract the changes for setting up code coverage into a standalone pull request.

Also I don't see coveralls results in the checks, do we need to enable loopback-connector-mongodb at coveralls side too?

hacksparrow: Removed all code coverage stuff.

test/objectid.test.js
```diff
@@ -57,19 +56,75 @@ describe('ObjectID', function() {
...
+    );
+
+    beforeEach(function(done) {
+      Article.deleteAll(done);
+    });
+
+    it('should throw if value is not an ObjectID-like string', async function() {
       try {
-        await Book.create({xid: 'x'});
+        await await Article.create({xid: '', title: 'abc'});
```
jannyHou: dup `await` here

Issue #499, opened 2019-03-22
How to escape special characters in Passwords to connect to MongoDB


arukris-cdw:
Hi Team, 

I am trying to connect to mongodb using loopback-connector-mongodb. When I start my loopback application, it throws the below error. My password will have characters like !$@ etc.. 

I tried to encode this characters like using %40 for @, %21 for ! and %24 for $. If my password has only @ character, the application is able to recognize %40 and start, but it throws the below error if my passwords has $ or ! . One of my service account has all these characters and I am having difficulties to start the application. Can you please let me know how to fix this problem. 

**Datasource:** 
{
    "quotesDS": {
    "host": "",
    "port": "",
    "url": "mongodb://<user>%40CORP.CDW.COM:<pwd>@host1,host2/QuotesDB?w=majority&wtimeoutMS=5000&replicaSet=RSDEV01&readPreference=primary&ssl=true&slaveOk=true&gssapiServiceName=mongodb&authMechanism=GSSAPI",
	"option": { "useNewUrlParser": true },
    "database": "QuotesDB",
    "name": "quotesDS",
    "connector": "mongodb",
    "replicaSet": "RSDEV01",
    "ssl": true
  }
}

**Error:**
C:\Projects\mongodb\mongocloud>node .
Connection fails: Error: Password contains an illegal unescaped character
It will be retried for the next request.
Connection fails: Error: Password contains an illegal unescaped character
It will be retried for the next request.
C:\Projects\mongodb\mongocloud\node_modules\loopback\lib\application.js:246
    throw err;
    ^

Error: **Cannot create data source "quotesDS": Cannot initialize connector "mongodb": Password contains an illegal unescaped character**
    at parseConnectionString (C:\Projects\mongodb\mongocloud\node_modules\mongodb\lib\url_parser.js:298:13)
    at parseHandler (C:\Projects\mongodb\mongocloud\node_modules\mongodb\lib\url_parser.js:129:14)
    at module.exports (C:\Projects\mongodb\mongocloud\node_modules\mongodb\lib\url_parser.js:25:12)
    at deprecated (internal/util.js:47:15)
    at connect (C:\Projects\mongodb\mongocloud\node_modules\mongodb\lib\operations\mongo_client_ops.js:188:3)
    at connectOp (C:\Projects\mongodb\mongocloud\node_modules\mongodb\lib\operations\mongo_client_ops.js:292:3)
[...]

PR #527, opened 2019-06-18
fix: support special characters in username and password

hacksparrow:
### Description

Support special characters in username and password.

#### Related issues

https://github.com/strongloop/loopback-connector-mongodb/issues/499
https://github.com/strongloop/loopback-connector-mongodb/pull/524

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -104,7 +104,7 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
 
   s.safe = s.safe !== false;
   s.w = s.w || 1;
-  s.url = s.url || generateMongoDBURL(s);
+  s.url = s.url ? encodeMongoDBURL(s.url) : generateMongoDBURL(s);
   s.useNewUrlParser = s.useNewUrlParser !== false;
   dataSource.connector = new MongoDB(s, dataSource);
   dataSource.ObjectID = mongodb.ObjectID;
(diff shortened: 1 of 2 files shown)


test/mongodb.test.js
```diff
@@ -3200,6 +3200,38 @@ describe('mongodb connector', function() {
     });
   });
 
+  describe('Test encodeMongoDBURL function', function() {
+    const module = require('../');
+
+    context('mongodb protocol', function() {
+      it('when username:password is provided', function() {
+        const url = 'mongodb://@$/! +=*%:@$/! +=*%@host:999/db';
```
bajtos: This URL is ambiguous. It can be parsed as `{user: '@$/! +=*%', password: '@$/! +=*%'}` but also as `{user: '@$/! +=*%:@$/! +=*%', password: undefined'}` or even `{user: '', password: undefined, host: '@$/! +=*%:@$/! +=*%@host'}`.

lib/mongodb.js
```diff
@@ -120,6 +120,45 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
...
 
+exports.encodeMongoDBURL = encodeMongoDBURL;
+
+function encodeMongoDBURL(url) {
+  let encodedUrl = url;
+  const atIndex = encodedUrl.lastIndexOf('@');
+  let protocol = 'mongodb://';
+  const srvProtocol = 'mongodb+srv://';
+
+  if (url.startsWith(srvProtocol)) protocol = srvProtocol;
```
bajtos: Why are you switching from `mongodb+src://` to `mongodb`? Those two protocols are different and there are good reasons why people want to use `mongodb+src://`.

hacksparrow:
> It's the responsibility of the person providing the URL string to ensure it's a valid URL and that all parts are correctly URLEncoded.

Agreed. We don't need this PR actually, just better documentation.

mercale:
> > It's the responsibility of the person providing the URL string to ensure it's a valid URL and that all parts are correctly URLEncoded.
> 
> Agreed. We don't need this PR actually, just better documentation.

  

> > It's the responsibility of the person providing the URL string to ensure it's a valid URL and that all parts are correctly URLEncoded.
> 
> Agreed. We don't need this PR actually, just better documentation.

I agree with this position, however any chance this can be reopen, because even after URL encoding the connection string, and adding the attribute   useNewUrlParser: true, the problem didn't go away, at least in my case it didn't help.  I am hoping that it is something that I am doing wrong, and jut need someone to point that out to me . Apologies for creating yet another issue on this regard.  Please help.  
image

PR #531, opened 2019-06-27
Callback with url parsing error

raymondfeng:
### Description

We should report errors for url parsing.

#### Related issues

https://github.com/strongloop/loopback-next/issues/3248#issuecomment-506390270

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -300,33 +300,43 @@ MongoDB.prototype.connect = function(callback) {
       }
     });
     debug('Valid options: %j', validOptions);
+
+    function onError(err) {
+      /* istanbul ignore if */
+      if (self.debug) {
+        g.error(
+          '{{MongoDB}} connection is failed: %s %s',
+          self.settings.url,
+          err
+        );
+      }
+      if (callback) callback(err);
+    }
+
     new mongodb.MongoClient(self.settings.url, validOptions).connect(function(
       err,
       client
     ) {
-      if (!err) {
-        if (self.debug) {
-          debug('MongoDB connection is established: ' + self.settings.url);
-        }
-        self.client = client;
-        // The database name might be in the url
-        return urlParser(self.settings.url, self.settings, function(err, url) {
-          self.db = client.db(
-            url.dbName || self.settings.database,
-            url.db_options || self.settings
-          );
-          if (callback) callback(err, self.db);
-        });
-      } else {
-        if (self.debug || !callback) {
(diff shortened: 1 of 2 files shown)


lib/mongodb.js
```diff
@@ -304,29 +304,40 @@ MongoDB.prototype.connect = function(callback) {
       err,
       client
     ) {
-      if (!err) {
+      if (err) {
+        if (callback) callback(err);
```
jannyHou: I think `if (callback) callback(err);` should be after the debug log `if (self.debug) { }`

test/mongodb.test.js
```diff
@@ -64,6 +64,19 @@ describe('lazyConnect', function() {
...
 
+  it('should report url parsing error', function(done) {
+    const ds = global.getDataSource({
+      url: 'mongodb://xyz:@127.0.0.1:4/xyz_dev_db',
+    });
+
+    ds.on('error', function(err) {
+      should.exist(err);
+      err.name.should.equal('MongoNetworkError');
+      err.message.should.match(/failed to connect to server/);
```
jannyHou: would [this code](https://github.com/strongloop/loopback-connector-mongodb/pull/531/files#diff-567d120f151f51afa9605ebcf0f20106R329) include the url in the error message?

raymondfeng: The url is included in the error by MongoDB driver.

lib/mongodb.js
```diff
@@ -304,29 +304,40 @@ MongoDB.prototype.connect = function(callback) {
...
-          if (callback) callback(err, self.db);
-        });
-      } else {
-        if (self.debug || !callback) {
           g.error(
             '{{MongoDB}} connection is failed: %s %s',
             self.settings.url,
             err
           );
         }
```
bajtos: Have you considered extracting this long block of error-handling code into a shared function inside `connect`? 

```js
if (err) return onError(err);
self.client = client;
// The database name might be in the url
return urlParser(self.settings.url, self.settings, function(err, url) {
  if (err) onError(err);
  // success - carry on
}

function onError(err) {
  if (callback) callback(err);
  if (self.debug) {
    // ...
  }
}
```

Also is it a good idea to pass the error to callback before logging it? Wouldn't it produce confusing logs, where the caller prints its error information and he low-level details are printed only later? Personally, I would log the error first and call back second.

bajtos:
@raymondfeng would you like to finish & land this pull request?

PR #533, opened 2019-07-01
fix: preserve id on update

hacksparrow:
Leave the id property alone on update. 

#### Related issues

- connect to https://github.com/strongloop/loopback-next/issues/3267

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -1608,23 +1608,24 @@ MongoDB.prototype.update = MongoDB.prototype.updateAll = function updateAll(
 
   where = self.buildWhere(modelName, where, options);
 
-  delete data[idName];
-  data = self.toDatabase(modelName, data);
+  let updateData = Object.assign({}, data);
+  delete updateData[idName];
+  updateData = self.toDatabase(modelName, updateData);
 
   // Check for other operators and sanitize the data obj
-  data = self.parseUpdateData(modelName, data, options);
+  updateData = self.parseUpdateData(modelName, updateData, options);
 
   this.execute(
     modelName,
     'updateMany',
     where,
-    data,
+    updateData,
     {upsert: false},
     function(err, info) {
       if (err) return cb && cb(err);
 
       if (self.debug)
-        debug('updateAll.callback', modelName, where, data, err, info);
+        debug('updateAll.callback', modelName, where, updateData, err, info);
 
       const affectedCount = info.result ? info.result.n : undefined;
 
--- a/test/mongodb.test.js
+++ b/test/mongodb.test.js
@@ -1032,6 +1032,27 @@ describe('mongodb connector', function() {
   });
 
   describe('updateAll', function() {
+    it('should not mutate the input data object', async function() {
+      const user = await User.create({name: 'Al', age: 31, email: 'al@strongloop'});
+      const userId = user.id;
+      const userData = user.toObject();
+      userData.age = 100;
+
+      await User.update(userData);
+      userData.should.have.property('id', userId);
+    });
+
+    it('should not mutate the input model instance', async function() {
+      const user = await User.create({name: 'Al', age: 31, email: 'al@strongloop'});
+      const userId = user.id;
+      user.age = 100;
+      user.name = 'Albert';
+
+      await User.update(user);
+      user.should.have.property('id', userId);
+      user.should.have.property('name', 'Albert');
+    });
+
     it('should update the instance matching criteria', function(done) {
       User.create({name: 'Al', age: 31, email: 'al@strongloop'}, function(
         err1,


hacksparrow:
I wonder why `delete data[idName];` was there in the first place. Maybe a remnant of copy-pasting from the implementation of other methods.

Do we need to add tests for this?

lib/mongodb.js
```diff
@@ -1606,8 +1606,6 @@ MongoDB.prototype.update = MongoDB.prototype.updateAll = function updateAll(
   const idName = this.idName(modelName);
 
   where = self.buildWhere(modelName, where, options);
-
-  delete data[idName];
```
raymondfeng: I think we should clone `data`, excluding `id`.

hacksparrow: Why clone? `data` could be deeply nested, and similar operations are performed in other places. It will introduce a new level of complexity and add to the scope.

`data` is passed around multiple layers and maybe be referenced from a controller, for example in the reported issue.

raymondfeng: For update to an existing document in mongodb, we delete the id on purpose IIRC. The current code changes the original input as complained by https://github.com/strongloop/loopback-next/issues/3267. I think the fix should leave `data` intact, but remove `id` from the object to be written to the database.

hacksparrow: @bajtos @b-admike @jannyHou thoughts?

Deleting the `id` is ideal and works perfectly, unless `data` is referenced in a controller. In the controller `data` exists as an instance of the Model, but it is the same object.

So, I guess we should go ahead with deep copy of `data` without the `id`.

raymondfeng: If we only have to remove `id`, shadow copy is probably good enough.

test/mongodb.test.js
```diff
@@ -1032,6 +1032,16 @@ describe('mongodb connector', function() {
...
 
   describe('updateAll', function() {
+    it('should preserve the id', async function() {
+      let user = await User.create({name: 'Al', age: 31, email: 'al@strongloop'});
+      const userId = user.id;
+      user = user.toObject();
+      user.age = 100;
+
+      await User.update(user);
+      user.should.have.property('id', userId);
```
bajtos: What is the expected `id` type - is it a string or `ObjectID` instance?

Also, `user` is the object given to `User.update`. Usually, methods don't modify objects passed in arguments, therefore I would expect that `user.id` remains the same after `User.update` as it was before making the call.

I.e. this test can be simplified as follows:

```ts
const userId = user.id;
user = user.toObject();
user.should.have.property('id', userId);
```

I think you should check `id` property in the value returned by `User.update` instead?

```ts
const result = await User.update(user);
result.should.have.property('id', userId);
```

The data returned by DAO methods like `update` don't necessarily match the data stored in the database. To be safe, you may want to check database records too.

```ts
const stored = await User.findById(userId);
stored.should.have.property('id', userId);
```

test/mongodb.test.js
```diff
@@ -1032,6 +1032,16 @@ describe('mongodb connector', function() {
   });
 
   describe('updateAll', function() {
+    it('should preserve the id', async function() {
```
bajtos: IIUC, this test is verifying that when the `id` is provided as string (e.g. when coming from the HTTP request body), then the connector is correctly coercing it to `ObjectID` value. Please capture this important information in the test name.

You may want to add another test to verify what happens when properties are modified directly in the model instance.

```ts
const user = await User.create({name: 'Al', age: 31, email: 'al@strongloop'});
const userId = user.id;
user.age = 100;

 await User.update(user);
// ...
```

test/mongodb.test.js
```diff
@@ -1032,6 +1032,16 @@ describe('mongodb connector', function() {
...
+      let user = await User.create({name: 'Al', age: 31, email: 'al@strongloop'});
+      const userId = user.id;
+      user = user.toObject();
```
bajtos: Let's call this property `userData` to make it clear that it's a plain-data object, not a model instance. (And also to avoid a mutable variable `user`.)

```suggestion
      userData = user.toObject();
```

test/mongodb.test.js
hacksparrow: I don't know about others, but this is the behavior of the Mongo connector on doing `User.update(user)`.

1. Modifies `user`
2. Returns an object with a `count` property, which is the number of items affected, does not return the model or a `toObject()` representation of it.

https://github.com/strongloop/loopback-next/issues/3267 is the result of modifying `user`.

`user` is not modified if it is an actual instance of a Model. However, `repository.create()` returns a `toObject()` version of the model, which the user passes to `repository.update()`, hence the modification.

test/mongodb.test.js
hacksparrow: > ... then the connector is correctly coercing it to ObjectID value

Nah, this test is just ensuring the `id` property is not deleted.

> You may want to add another test to verify what happens when properties are modified directly in the model instance.

Isn't it already happening on the model instance? 😃

test/mongodb.test.js
hacksparrow: That makes we wonder why should `repository.create()` not return the model instance?

bajtos: > Returns an object with a count property, which is the number of items affected, does not return the model or a toObject() representation of it.

Ah, good point, I never remember which method returns `{count}` and which returns the updated model instance. Thanks for clarifying!

> Modifies `user` if it is not an instance of a Model.

While `user` parameter can be modified inside a method, this change is not visible to the caller.

```js
const x = {orig: true};
function f(data) { data = {modified: true}; }
f(x);
console.log(x);
// prints: { orig: true }
```

test/mongodb.test.js
bajtos: > > You may want to add another test to verify what happens when properties are modified directly in the model instance.
>
> Isn't it already happening on the model instance?

No, you are making the change on the object returned by `user.toObject()`. That's a new object instance with prototype set to `Object`, it's not the model instance with prototype set to `User`.

hacksparrow: Updated.

test/mongodb.test.js
```diff
@@ -1032,6 +1032,27 @@ describe('mongodb connector', function() {
...
+      user.age = 100;
+
+      await User.update(user);
+      user.should.have.property('id', userId);
+    });
+
+    it('should reflect the changes in the model object', async function() {
```
raymondfeng: The test name is misleading. IIUC, it is testing `update() does not mutate the input model object`.

test/mongodb.test.js
bajtos: @hacksparrow this discussion thread is not resolved.

See also Raymond's comment below:

> The test name is misleading. IIUC, it is testing `update() does not mutate the input model object`.

I think you need to decide what is the scenario you are testing and then rework & clarify the tests to test only that scenario and nothing else.

hacksparrow:
@bajtos @raymondfeng looks good now?

test/mongodb.test.js
```diff
@@ -1032,6 +1032,27 @@ describe('mongodb connector', function() {
...
+      let user = await User.create({name: 'Al', age: 31, email: 'al@strongloop'});
+      const userId = user.id;
+      user = user.toObject();
+      user.age = 100;
+
+      await User.update(user);
+      user.should.have.property('id', userId);
+    });
+
+    it('should not mutate the input model instance', async function() {
```
b-admike: Maybe I misunderstand, but the test name is saying it _shouldn't_ mutate input model instance, but we do an update which does change the instance? Can you please clarify the use case?

hacksparrow: @bajtos what are your thoughts? Since you suggested the name.

bajtos: Here is the essence of the test:

```ts
const user = await User.findById(someId);
user.name = 'new name';
await User.update(user);
```

In the test, we are verifying that `user` was not changed by the operation (e.g. the `id` property was not deleted).

Of course, the data stored in the database is modified by the operation.

I proposed to use names "input data object" and "input model instance" to describe content of the variable passed in `User.update` arguments.

@b-admike can you suggest a better naming scheme?

b-admike: Ah okay, makes sense. I get that by `mutate` we mean we don't change the model instance in-place but do a copy instead. I think the name is good as it is, thank you for the explanation.

test/mongodb.test.js
bajtos: I am afraid my comment was not resolved :(

hacksparrow:
@bajtos errr thanks for pointing our. Updated.

PR #546, opened 2019-10-25
Implicit null type

louisbeullens:
### Description

added implicitNullType boolean to connector.settings

default false to prevent breaking changes

when true it allows the connector not to replace null by {$type:10} in where clause

### Checklist


slnode:

PR #550, opened 2019-11-01
ObjectID Coercion supports array of values

### LINKED ISSUE #545: `coercePropertyValue` Does Not Account For Array of ObjectID or Array of ObjectID strings
Arrays of ObjectID or ObjectId-like-strings cannot be saved/updated/created and attempting to do so results in an error.

1. Create any Entity/Model.
2. Decorate a model w/ `{strictObjectIDCoercion:true}`.
2. Decorate a property with either:
```ts
@property.array(String, {mongodb: {dataType:'ObjectID'}})
```
OR
```ts
@property({
     type:'array',
     itemType: 'string',
     mongodb: {
        dataType: 'ObjectID',   
    }
})
```
I've tried many other variations. No matter what, `coercePropertyValue` in `mongodb.js` does not account for the fact that the `propValue` might be an Array *unless* it is a decimal128 array rather than an Array of ObjectID or ObjectID-like-strings.

# Expected result
The expectation is that one should be able to simply add an ObjectId string (or even ObjectId  - this results in an error as well) to the array and update/save/create the data.

# Notes
- Also attempting to persist a property with an empty array (that is decorated with any of the above property decorators) will result in an error due to coercePropertyValue continuing down the same code path and trying to evaluate the property value.
- I am mainly submitting this issue to make the owners aware of the problem and to hopefully aid a traveler in search of an answer or information related to this problem.

Tested on version: 5.0.1

mitsos1os:
### Description

When a property is of type array of 'ObjectId', it now properly is coerced.

#### Related issues

- fixes #545 

### Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -1965,7 +1965,7 @@ function coerceToObjectId(modelCtor, propDef, propValue) {
  */
 function isObjectIDProperty(modelCtor, propDef, value, options) {
   if (!propDef) return false;
-
+  if (Array.isArray(value)) return isObjectIDProperty(modelCtor, propDef, value[0], options);
   if (typeof value === 'string' && value.match(ObjectIdValueRegex)) {
     if (isStoredAsObjectID(propDef)) return true;
     else return !isStrictObjectIDCoercionEnabled(modelCtor, options);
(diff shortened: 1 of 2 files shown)


slnode:

dhmlau:
@mitsos1os, thanks for your PR. Could you please rebase your branch? Thanks!

mitsos1os:
Hi @dhmlau , I have rebased and also squashed the commits.
However I came across a bug, that for some reason did not fail the tests earlier.
When using the notation of a type with the extra `mongodb` parameter such as in my tests:
```js
const Article = ds.createModel(
      'ArticleC',
      {
        xid: {type: String, mongodb: {dataType: 'objectid'}},
        xidArr: {type: [String], mongodb: {dataType: 'objectid'}},
        title: String,
      }
    );
```
Although the connector successfully saves the model in the database as an array of ObjectIds, when retrieving the model through `find` then because it is an array of values it is instantiated as a `List` using its declared type *(which is String in this case)*, not taking into account the `mongodb` parameter of the type definition.

This takes place here: https://github.com/strongloop/loopback-datasource-juggler/blob/d54d76947724d75a0f46b151c064742c678b0bbe/lib/model.js#L357

How should we handle this?

mitsos1os:
Any news for the PR here?

dhmlau:
@mitsos1os, my apologies. I just realized your PR is similar to the PR we just merged: https://github.com/strongloop/loopback-connector-mongodb/pull/560.  What do you think?

mitsos1os:
> @mitsos1os, my apologies. I just realized your PR is similar to the PR we just merged: #560. What do you think?

@dhmlau No problem. It fixes the same thing so I will close mine. I know you must have your hands full but it would be great if we could tackle an open PR a bit sooner since there is small number of them open, which makes it quite manageable.

Anyway, thanks for the  info!

PR #553, opened 2019-11-21
fix creation of LB4 models with auto-generated id

bajtos:
- Add a try/catch block to prevent `coerceId` from crashing the process on uncaught exception (forward the coercion error via callback).

- Fix `isObjectIDProperty` to use the correct ObjectID constructor in `instanceof` check.

This fixes the problem described in https://github.com/strongloop/loopback-next/issues/4059#issuecomment-557138697.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-mongodb) 👈

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -596,7 +596,13 @@ MongoDB.prototype.create = function(modelName, data, options, callback) {
       return callback(err);
     }
     idValue = result.ops[0]._id;
-    idValue = self.coerceId(modelName, idValue, options);
+
+    try {
+      idValue = self.coerceId(modelName, idValue, options);
+    } catch (err) {
+      return callback(err);
+    }
+
     // Wrap it to process.nextTick as delete data._id seems to be interferring
     // with mongo insert
     process.nextTick(function() {
@@ -1963,7 +1969,7 @@ function isObjectIDProperty(modelCtor, propDef, value, options) {
   if (typeof value === 'string' && value.match(ObjectIdValueRegex)) {
     if (isStoredAsObjectID(propDef)) return true;
     else return !isStrictObjectIDCoercionEnabled(modelCtor, options);
-  } else if (value instanceof ObjectID) {
+  } else if (value instanceof mongodb.ObjectID) {
     return true;
   } else {
     return false;
--- a/test/objectid.test.js
+++ b/test/objectid.test.js
@@ -126,5 +126,23 @@ describe('ObjectID', function() {
       const found = await Article.findOne({where: {title: 'abc'}});
       found.xid.should.be.an.instanceOf(ds.ObjectID);
     });
+
+    it('handles auto-generated PK properties defined in LB4 style', async () => {
+      const Note = ds.createModel('NoteLB4', {
+        id: {
+          type: 'string',
+          id: true,
+          generated: true,
+          mongodb: {dataType: 'ObjectID'},
+        },
+        title: {
+          type: 'string',
+          required: true,
+        },
+      });
+
+      const result = await Note.create({title: 'hello'});
+      // the test passes when this call does not throw
+    });
   });
 });


agnes512:
I checked all the mongo id test cases, I try to summarize the CRUD behavior on  `strictObjectIDCoercion` and `dataType: ObjectID` cause I didn't find any documentation to describe it:

|        | `strict`::x:`type`::x: | `strict`::heavy_check_mark: `type`: :x: | `strict`: :x: `type`:  :heavy_check_mark:  | `strict`:  :heavy_check_mark: `type`: :heavy_check_mark:|
| :--------------: |:-------------:| :-------------:|:-------------:|:------------------:|
| create with `ObjectId` | T       |    unknown   |   unknown    |  T     |
| create with `objId-like-str` | T | T       |    unknown    | T , returns `ObjectID`|
| create with invalid `str` |   T |  F| unknown| F |
| find with`ObjectId` | T   |  T , returns `str` id| unknown| T, `type` overrides, returns `ObjectID` |
| find with `objId-like-str`| T   |    unknown | unknown|  T , bc `strict` is true |
| update with `ObjectId` | T | unknown| unknown| T bc`type` is set |
| update with `objId-like-str`| T      |    unknown | unknown|T bc`type` is set |

Please correct me if anything is incorrect.

bajtos:
@agnes512 It's a great idea to document behavior for different combinations of `strictObjectIDCoercion` and `dataType: ObjectID`, maybe we can add this table to connector's README file?

To be honest, I find it difficult to understand the values in table cells. What does it mean when the value is `T`,  `F` or `unknown`?

I think that my pull request should not be changing the behavior, I am just fixing an incorrectly written `instanceof` check that prevented the `create` operation from returning data back to the user.

lib/mongodb.js
```diff
@@ -1963,7 +1969,7 @@ function isObjectIDProperty(modelCtor, propDef, value, options) {
   if (typeof value === 'string' && value.match(ObjectIdValueRegex)) {
     if (isStoredAsObjectID(propDef)) return true;
     else return !isStrictObjectIDCoercionEnabled(modelCtor, options);
-  } else if (value instanceof ObjectID) {
+  } else if (value instanceof mongodb.ObjectID) {
```
bajtos: `ObjectID` is a factory function that creates `new mongodb.ObjectID`, it's not a class. Therefore the expression `value instanceof ObjectID` is never true.

test/objectid.test.js
```diff
@@ -126,5 +126,23 @@ describe('ObjectID', function() {
...
       found.xid.should.be.an.instanceOf(ds.ObjectID);
     });
+
+    it('handles auto-generated PK properties defined in LB4 style', async () => {
+      const Note = ds.createModel('NoteLB4', {
+        id: {
+          type: 'string',
+          id: true,
+          generated: true,
+          mongodb: {dataType: 'ObjectID'},
```
bajtos: This is the property definition used in LB4.

When the primary keys is auto-generated by the database, juggler replaces the type provided by the user (`string`) with the type defined by the connector (`mongodb.ObjectID`).

As a result, we end up with a property defined as `{type: mongodb.ObjectID, mongodb: {dataType: 'ObjectID'}}`, which is something we were not handling well before my pull request.

agnes512:
> To be honest, I find it difficult to understand the values in table cells. What does it mean when the value is T, F or unknown?

In order to keep it short I use short terms. T for `true` and F for `false`. 
For example, when `strictObjectIDCoercion` sets to true and `dataType: ObjectID` is defined ( the top right case), it is true that it can be created with `ObjectId`.

> I think that my pull request should not be changing the behavior

The top right case was supposed to be true, but we didn't have a test case, and it was a bug. I think your PR fix it 👍

PR #591, opened 2020-08-13
SPIKE: New objectid

hacksparrow:
## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-mongodb) 👈

- [ ] `npm test` passes on your machine


SPIKE.md
```diff
@@ -0,0 +1,95 @@
...
+- Changes in `loopback-datasource-juggler/lib/dao.js`:
+  - Coercion has to be applied in DAO for some queries.
+  - Depending on the query, there may be different restrictions and rules.
+
+In the time assigned for this spike I could find out the behavior of all the query methods. Also, I didn't have enough time to explore in deep how LB4 handles `mongodb: {dataType: 'ObjectID'}`. I experimented with a simple app with the changes in `loopback-connector-mongodb/lib/mongodb.js` and `loopback-datasource-juggler/lib/dao.js` but was hit with `Type 'MongoDataSource' has no properties in common with type 'LifeCycleObserver'.`; I wonder why it would happen.
+
+## Implementation proposal
+
+- Database-specific connectors should provide their coercion function to DAO.
+- DAO should apply the coercion function on the result object and return it to the user without changing the Model's actual properties.
```
bajtos, line 90: Have you considered this simpler implementation?

- In code that's handling property initialization (see [ModelBaseClass.prototype._initProperties](https://github.com/strongloop/loopback-datasource-juggler/blob/261dd1c865e8b197882a161b909b90b0e6c3c2a3/lib/model.js#L202-L207)), we can covert property values provided by the caller to the types defined in model definition.
- ObjectID types are defined as `{type: 'string', mongodb: {dataType: 'ObjectID'}}`, therefore they will be converted to a string (as we want).
- No extra connector-specific coercion rules are required.

This was one of the reasons why I choose the design where the property `type` is a well-known primitive like a `string` or a `number`, and database-specific information (e.g. how to represent/store the property in the database) is nested under connector-specific config. As long as the database-specific value types play well with JavaScript's built-in coercion (to a string, to a number), no extra work is needed in juggler to support these types.

Consider the "bit" example you described earlier (`{type: 'number', mysql: {dataType: 'bit'}}`). Even if the mysql connector (or driver) returns the property value as a boolean, the generic coercion algorithm will attempt to convert the value to a number, which works exactly as we want.

```
$ node
> +true
1
> +false
0
```

Having said that, there is also the argument that we want to hide the complexity of database-specific types like `ObjectID` from juggler, keep it contained inside the connector only. In that case, all connectors methods should accept stringified object ids and also return stringified object ids. The task of processing model definitions and applying any necessary coercion becomes the responsibility of the connector. 

I believe we have this approach already in place, to support `decimal128` datatype in MongoDB connector.

This solution has an advantage that very little (if any) changes in juggler are necessary, thus we are not going to break existing users of other connectors. It will also make our life easier in the future, when we implement a LB4 `Repository` class that's calling connectors directly, removing juggler from the picture.

Personally, I find this second approach better for the longer term.

Either way, please update your proposal to take into account the existing code for `decimal128` and come up with a consistent solution for both `ObjectID` and `decimal128` types. IMO, these types should be handled the same way.

test/objectid.test.js
```diff
@@ -101,12 +101,12 @@ describe('ObjectID', function() {
...
 
-  context("mongodb: {dataType: 'objectid'}", function() {
+  context("mongodb: {dataType: 'objectID'}", function() {
     const Article = ds.createModel(
       'ArticleC',
       {
-        xid: {type: String, mongodb: {dataType: 'objectid'}},
-        xidArr: {type: [String], mongodb: {dataType: 'objectid'}},
+        xid: {type: String, mongodb: {dataType: 'objectID'}},
+        xidArr: {type: [String], mongodb: {dataType: 'objectID'}},
```
bajtos, line 109: Let's support all three casing variants please: `ObjectID` (LB4 style), `objectid` and `objectID`.

bajtos (changes requested):
Thank you @hacksparrow for a detailed description of the proposal and options you rejected along the way.

I'd like to discuss how & where to coerce ObjectID values coming from the database into string values that should be stored in model instance properties, see below.

bajtos:
Few more thoughts: we already have useful helpers in place to help with coercion of property values - see [`visitAllProperties`](https://github.com/strongloop/loopback-connector-mongodb/blob/36c4f10abd22af9c05b9eac47fb8d8356ae11a1c/lib/mongodb.js#L2176) and [`coercePropertyValue`](https://github.com/strongloop/loopback-connector-mongodb/blob/36c4f10abd22af9c05b9eac47fb8d8356ae11a1c/lib/mongodb.js#L2204). What's important, these helpers are already taking into account different ways how a model property can be defined, for example in a nested (embedded) model - see [`nested decimal props` tests](https://github.com/strongloop/loopback-connector-mongodb/blob/36c4f10abd22af9c05b9eac47fb8d8356ae11a1c/test/decimal.test.js#L107).

If we decide to move the coercion logic to juggler, then we must support nested properties too.

If we decide to keep the coercion logic inside the connectors (which I prefer), then it would be great to use this opportunity to extract these two helpers into `loopback-connector` and refactor them to be usable in other connectors beside MongoDB. Ideally, the shared logic for visiting all properties and obtaining their definition should be inside `loopback-connector` and the MongoDB connector should deal only with connector-specific rules - ow to convert `dataType: 'ObjectID'` and `dataType: 'decimal128'`.

test/objectid.test.js
jannyHou: +1 for case insensitive

hacksparrow:
Closing in favor of https://github.com/strongloop/loopback-connector-mongodb/pull/596. #596 is much simpler.

PR #596, opened 2020-09-03
Alt SPIKE: New objectid

hacksparrow:


lib/mongodb.js
```diff
@@ -348,9 +348,10 @@ MongoDB.prototype.getTypes = function() {
   return ['db', 'nosql', 'mongodb'];
 };
 
-MongoDB.prototype.getDefaultIdType = function() {
-  return ObjectID;
-};
+// SPIKE: get rid of this. TODO: remove other redundant code too
+// MongoDB.prototype.getDefaultIdType = function() {
+//   return ObjectID;
+// };
```
bajtos, line 354: I think this change will break compatibility with LB 3.x. Which is fine, we just need to clearly document it and remove `juggler@3` from the test suite.

As part of the doc changes, we should add a "Long Term Support" section to README (see [`strongloop/loopback`](https://github.com/strongloop/loopback#module-long-term-support-policy) for inspiration) and also explain which connector versions are compatible with which LoopBack versions.

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+- [Approaches](#approaches)
+- [Developer Notes](#developer-notes)
+- [Implementation Proposal](#implementation)
+- [Follow-up task](#follow-up-task)
+
+## The Problem
+
+`ObjectID` is a native MongoDB datatype, which is used as the datatype of its primary key (_id_) field; other fields
+can also be of `ObjectID` type.
+
```
bajtos, line 18: Let's add a short paragraph to explain that:

MongoDB distinguishes between id values that are a `string` and an `ObjectID`. A query using a `string` value does not match records having `ObjectID` value and vice versa. This is especially important for foreign keys, e.g. when `Order` property `customerId` is referencing an `id` of a `Customer`, then both properties must always hold `ObjectID` values. If one of the property (typically `customerId`) is stored as a `string`, then LB relations stop working.

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+In the current version of Juggler, the interpretation of `ObjectID` field is a sub-optimal experience supported by a complex
+determination process in an already complex codebase.
+
+Examples of bad experiences: 
+
+1. Fields need not be set to `mongodb: {dataType: 'ObjectID'}` to be interpreted as `ObjectID`.
```
bajtos, line 24: I find this sentence a bit difficult to understand. Are you trying to say this? 👇🏻 

```suggestion
1. Model properties must be defined as `{type: 'string', mongodb: {dataType: 'ObjectID'}}` to be interpreted as `ObjectID`.
```

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+tried searching the database by that string, because it is not a string any more.
+3. To add to the confusion, this behavior can be configured at model and property levels.
+
+Complexity is sometimes unavoidable, but a less complex state is always the desired state. It creates better experience for
+the consumers, and makes development and maintenance a more pleasant experience for the creators.
+
+## Proposed Solution
+
+"Do not interpret any property not set with `mongodb: {dataType: 'ObjectID'}` as an `ObjectID`. Use string values for `ObjectID`
+in public APIs and manage the coercion processes for the user."
```
bajtos, line 36: > Do not interpret any property not set

In my experience, double negatives are difficult to understand, especially for native English speaker. Let's rephrase this in a positive way (what we want, instead what we do not want).


```suggestion
Database-specific types like `ObjectID` and `Decimal128` should be considered as an implementation detail of a database connector. Values of these types must not leak from the connector to juggler and user code, code outside the connector should see a database-agnostic type like a `string`. The connector must always convert incoming values (property values, filtering queries) from database-agnostic type to database-specific type on input, and convert database-specific types back to database-agnostic type on output.

Specifically for `ObjectID`:

- Properties that should be stored as `ObjectID` in MongoDB must be defined as `{type: 'string', mongodb: {dataType: 'ObjectID'}}`.
- There will be no auto-magic coercion from `string` to `Object` for other properties (those defined without `dataType: 'ObjectID'`).
```

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+    dataType: 'bit',
+    length: 1,
+  }
+}
+```
+
+### 2. mongodb: {dataType: 'objectID'}
+
+Setting a property to `mongodb: {dataType: 'ObjectID'}`, as it is already being done, is the better approach for marking the 
+property as an `ObjectID`.
```
bajtos, line 71: Can you please mention that we are already using the same approach for data type `decimal128`?

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+## Proof of concept
+
+The tests in `test/new-objectid.test.js` is a demonstration of the behavior of the new `mongodb: {dataType: 'ObjectID'}` property.
+
+## Implementation
+
+All changes are limited to `loopback-connetor-mongodb` only, Juggler is not affected.
+
+- Before querying database:
+  - Convert `id` value to `ObjectID`.
```
bajtos, line 82: Is it possible to use an id value that's not an `ObjectID`, e.g. a number? I vaguely remember that in such case, LoopBack 3 may be using a new property `_id` to hold the actual database id (e.g. `{_id: objectIdGeneratedByMongo, id: 123}`).

Can you please verify?

This item caught my attention because I find it a bit odd that we have to treat `id` (primary key) values differently from other properties, considering that both primary keys and other properties must be defined the same way (`mongodb: {dataType: 'ObjectID'}`).

Also if we have special treatment for primary keys, then shouldn't we also apply the same rules to foreign keys? 

Have you considered a simpler design, where we apply the same rules for all properties, regardless of whether they are a primary key or not? Is there any edge case that not work with that solution?

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+  - Walk through the data object and convert all properties defined as `mongodb: {dataType: 'ObjectID'}`.
+- When receiving data from database
+  - Convert `id` value to `String`:
```
bajtos, line 85: Ditto - why do we need special handling of primary keys?

SPIKE.md
```diff
@@ -0,0 +1,92 @@
...
+  - Walk through the data object and convert all properties defined as `mongodb: {dataType: 'ObjectID'}` to `String`
+- Remove or refactor all helper functions and properties like `typeIsObjectId()`, `strictObjectIDCoercion` etc., related to previous behavior of `ObjectId`.
+- There are some pieces of code that can be refactored, which involves the new changes. I think we should refactor it now now, than accumulate tech debt.
```
bajtos, line 88: +1 to refactor and clean up related code.

IMO, we should find a way how to unify the (more recent) implementation of `Decimal128` coercion with the pre-historic coercion of `ObjectID`. 

From what I remember, I was guiding the implementation of `Decimal128` coercion to introduce helpers that can be used for other data types like `ObjectID` too, for example `visitAllProperties` and `hasDataType`. I think we should check how far we can get with reusing those helpers for `ObjectID`, and them make the necessary improvements to fix them to support any edge cases that don't work.

Also eventually, I would like to make some of these helpers database-agnostic and move them to `loo
pback-connector`, so that we can start using them in other connectors too.

lib/mongodb.js
bajtos: @hacksparrow can you please capture my comment above in the spike document?

lib/mongodb.js
```diff
@@ -401,6 +403,12 @@ MongoDB.prototype.fromDatabase = function(modelName, data) {
       if (data[p] instanceof mongodb.Binary) {
         // Convert the Binary into String
         data[p] = data[p].toString();
+      // SPIKE: properties that should be converted to string from ObjectID
+      // SPIKE: do the same for toDatabase
+      } else if (p === 'id' || prop.mongodb && prop.mongodb.dataType.match(/objectid/i)) {
```
bajtos, line 408: Please keep in mind that primary keys are not required to be called `id`. If you really need to identify primary keys here (which I think you should not need to), then you need to check if `prop.id` is set (I think). I also vaguely remember that `prop.id` can be a number to describe multi-column index (e.g. on GitHub, `userId` would have `id: 1` and `repoId` would have `id: 2` to define `(userId, repoId)` as the primary key). I am not sure if those numeric values are indexed from `0` or `1` though.

lib/mongodb.js
```diff
@@ -448,6 +458,16 @@ MongoDB.prototype.toDatabase = function(modelName, data) {
         coordinates: [data[p].lng, data[p].lat],
         type: 'Point',
       };
+    // SPIKE: Ideally coercePropertyValue should handle this but it does not have enough info
+    } else if (prop && prop.type === String) {
```
bajtos, line 462: Please try find a way how to use the existing code for converting `string` to `Decimal128`.

lib/mongodb.js
```diff
@@ -401,6 +403,12 @@ MongoDB.prototype.fromDatabase = function(modelName, data) {
       if (data[p] instanceof mongodb.Binary) {
         // Convert the Binary into String
         data[p] = data[p].toString();
+      // SPIKE: properties that should be converted to string from ObjectID
+      // SPIKE: do the same for toDatabase
+      } else if (p === 'id' || prop.mongodb && prop.mongodb.dataType.match(/objectid/i)) {
+        if (data[p]) {
+          data[p] = data[p].toString();
+        }
```
bajtos, line 411: Please try find a way how to use the existing code for converting `string` to `Decimal128`.

lib/mongodb.js
```diff
@@ -1508,7 +1531,8 @@ MongoDB.prototype.count = function count(modelName, where, options, callback) {
  */
 MongoDB.prototype.replaceById = function replace(modelName, id, data, options, cb) {
   if (this.debug) debug('replace', modelName, id, data);
-  const oid = this.coerceId(modelName, id, options);
+  // SPIKE: every id should be converted to ObjectID before query
+  const oid = ObjectID(id);
```
bajtos, line 1535: See my comments above. IMO, the conversion to ObjectID should be applied only if the id property is defined with data type ObjectID.

It would be great if we could re-use existing bits from the infrastructure for coercing arbitrary model properties.

test/new-objectid.test.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+  );
+
+  const Author = ds.createModel(
+    'Author',
+    {
+      name: String
+    }
+  );
+
```
bajtos, line 37: Please add few tests to verify how nested properties are handled.

See the tests we added for Decimal128 in fb41e40777 and 25122e02a59 for inspiration.

bajtos (changes requested):
You are making good progress 👏🏻 

I have few more comments to discuss, PTAL 👇🏻

SPIKE.md
hacksparrow: Nah, I am talking about the current behavior of interpreting any `ObjectID`-like string as an `ObjectID`.

SPIKE.md
hacksparrow: The special treatment for primary key is the existing behavior, let me update the implementation to apply a common rule for all.

SPIKE.md
hacksparrow: The special treatment for primary key is the existing behavior, let me update the implementation to apply a common rule for all.

SPIKE.md
hacksparrow: > Is it possible to use an id value that's not an ObjectID, e.g. a number?

Yes, but using non-ObjectID `_id`  it comes with unexpected behaviors, so not worth using it.

bajtos: > > Is it possible to use an id value that's not an ObjectID, e.g. a number?
> 
> Yes, but using non-ObjectID `_id` it comes with unexpected behaviors, so not worth using it.

I am not referring to non-ObjectID `_id` field. I mean a model that's using a number as the primary key on the LoopBack side.

Here is some code you can add to `lib/objectid.test.js` to better understand what's going on.

```js
  context.only('numeric id', function() {
    const User = ds.createModel(
      'UserNum',
      {
        id: {type: 'number', id: true, generated: false},
        email: {type: 'string'},
      },
      {strictObjectIDCoercion: true},
    );

    beforeEach(() => User.deleteAll());
 
    it('is supported and uses _id: ObjectID in background', async () => {
      console.log('id definition', User.definition.properties.id);

      const u = await User.create({id: 1, email: 'test@example.com'});
      console.log(u);
      console.log(u.toJSON());

      const found = await User.findById(u.id);
      console.log('found', found);
    });
  });
```

Console output:

```
id definition {
  type: [Function: Number],
  id: true,
  generated: false,
  updateOnly: false
}
{ id: 1, email: 'test@example.com' }
{ id: 1, email: 'test@example.com' }
found { email: 'test@example.com', id: 1 }
```

Interestingly enough, when I change `id` definition and remove `generated: false`, and don't provide any `id` value when creating a new user, then the connector somehow starts storing ObjectID values in the numeric `id` property.

```
id definition { type: [Function: Number], id: true, updateOnly: false }
{ email: 'test@example.com', id: 5f5b71894004d51b2a65de00 }
{ id: 5f5b71894004d51b2a65de00, email: 'test@example.com' }
found { email: 'test@example.com', id: NaN }
```

This looks like a bug to me, notice that the model instance returned by `findById` has `id: NaN`.

SPIKE.md
```diff
@@ -32,10 +33,15 @@ the consumers, and makes development and maintenance a more pleasant experience
...
 
-"Do not interpret any property not set with `mongodb: {dataType: 'ObjectID'}` as an `ObjectID`. Use string values for `ObjectID`
-in public APIs and manage the coercion processes for the user."
+Database-specific types like `ObjectID` and `Decimal128` should be considered as an implementation detail of a database connector.
+
+Values of these types must not leak from the connector to Juggler and user code, code outside the connector should see a database-agnostic type like a `string`. The connector must always convert incoming values (property values, filtering queries) from database-agnostic type to database-specific type on input, and convert database-specific types back to database-agnostic type on output.
+
+Specifically for `ObjectID`:
+- Properties that should be stored as `ObjectID` in MongoDB must be defined as `{type: 'string', mongodb: {dataType: 'ObjectID'}}`.
+- There will be no auto-magic coercion from `string` to `Object` for other properties (those defined without `dataType: 'ObjectID'`).
```
bajtos, line 42:
```suggestion
- There will be no auto-magic coercion from `string` to `ObjectID` for other properties (those defined without `dataType: 'ObjectID'`).
```

SPIKE.md
bajtos: I think you did not update the implementation yet, or at least I am not able to find those changes.

SPIKE.md
hacksparrow: @bajtos 

> Is it possible to use an id value that's not an ObjectID, e.g. a number? I vaguely remember that in such case, LoopBack 3 may be using a new property _id to hold the actual database id (e.g. {_id: objectIdGeneratedByMongo, id: 123}).

Yes, it possible to use an id value that's not an ObjectID, but the custom id field is not saved in the database, it is logically translated in the model from `_id` to whatever name was specified.

> when I change id definition and remove generated: false, and don't provide any id value when creating a new user, then the connector somehow starts storing ObjectID values in the numeric id property.

This happens on the master branch too. We should throw an error saying `id` is required, like how specifying the `id` for `generated: 1` throws an error.

Need some clarifications:
- If the model is defined as `{id: {id: true, type: 'string', generated: true }}`, should it be interpreted as `{id: {id: true, type: 'string', mongodb: {dataType: 'objectID'}}}`? After all, the autogenerated id will be an `ObjectID`.
- How should `{id: {id: true, type: 'number', generated: true }}` be treated? It is essentially saying `{id: {id: true, type: 'number', mongodb: {dataType: 'objectID'}}}`.

I have created a branch - https://github.com/strongloop/loopback-connector-mongodb/tree/non-objectid-id - for quickly testing the current behavior on `master`. File of interest: https://github.com/strongloop/loopback-connector-mongodb/blob/non-objectid-id/test/new-objectid.test.js.

hacksparrow: > This looks like a bug to me, notice that the model instance returned by findById has id: NaN.

Can we address this in a separate issue since the fix lies in Juggler? We will be able to make faster progress by not including changes in Juggler. From my initial look, it seems to be caused by the difference in behavior of the various DAO methods as [noted here](https://github.com/strongloop/loopback-next/issues/889#issuecomment-683887309).

bajtos: > Yes, it possible to use an id value that's not an ObjectID, but the custom id field is not saved in the database, it is logically translated in the model from `_id` to whatever name was specified.

Interesting, I thought the model would have two properties - user-defined `id` and database-generated `_id`. Thank you for taking a look and finding out what's the actual behavior.

> If the model is defined as `{id: {id: true, type: 'string', generated: true }}`, should it be interpreted as `{id: {id: true, type: 'string', mongodb: {dataType: 'objectID'}}}`? After all, the autogenerated id will be an `ObjectID`.

One option is to silently (auto-magically) assume `mongodb: {dataType: 'objectID'}`. A better solution may be to throw a descriptive error and asks users to specify `mongodb: {dataType: 'objectID'}` in the property definition.

> * How should `{id: {id: true, type: 'number', generated: true }}` be treated? It is essentially saying `{id: {id: true, type: 'number', mongodb: {dataType: 'objectID'}}}`.

IMO, `{type: 'number', mongodb: {dataType: 'objectID'}}` is not a valid property definition, similarly for `{type: 'date', mongodb: {dataType: 'objectID'}}`, etc.

I am proposing to throw an error when such property definition is detected.

Connectors can provide `define`, `defineProperty` and `defineForeignKey` functions that are called by juggler's `DataSource` object to notify the connector about new models/properties/foreign-keys defined by the application. I think it would be best to implement these methods in MongoDB connector to validate property definitions at setup (define) time, instead of deferring the validation until the first DAO method is called.

bajtos: > > This looks like a bug to me, notice that the model instance returned by findById has id: NaN.
> 
> Can we address this in a separate issue since the fix lies in Juggler? We will be able to make faster progress by not including changes in Juggler. From my initial look, it seems to be caused by the difference in behavior of the various DAO methods as [noted here](https://github.com/strongloop/loopback-next/issues/889#issuecomment-683887309).

Sure, that's fine. 

However, I think my proposal to reject property definitions like `{type: 'number', mongodb: {dataType: 'objectID'}}` would solve the issue and it is something we can implement inside the MongoDB connector.

hacksparrow: > I think it would be best to implement these methods in MongoDB connector to validate property definitions at setup (define) time, instead of deferring the validation until the first DAO method is called.

Let's do it in a follow up story. Let's focus specifically on `ObjectID`-to-string conversion and vice versa in this spike. If we include validation in the connector, it will further drag on the spike.

test/new-objectid.test.js
```diff
@@ -0,0 +1,257 @@
...
+      strId: String,
+      strIds: [String],
+      authorId: {type: [String], mongodb: {dataType: 'objectID'}}
+    }
+  );
+
+  const Author = ds.createModel(
+    'Author',
+    {
+      id: {id: true, type: String, mongodb: {dataType: 'objectID'}},
```
jannyHou, line 35: nitpick: according to "if specified with `{id: true}`, need not add `mongodb: {dataType: 'ObjectID'}` (it will be automatically inferred)" the mongodb config here can be omitted.

test/new-objectid.test.js
```diff
@@ -0,0 +1,257 @@
...
+    await Book.deleteAll();
+    await Author.deleteAll();
+    await Deep.deleteAll();
+
+    Author.hasMany('books');
+    Book.belongsTo('author');
+  });
+
+
+  // SPIKE: in these tests there is no way of knowing if objectID-like strings are actually stored
```
jannyHou, line 83: For the spike: we test the raw data from database for decimal properties like [this](https://github.com/strongloop/loopback-connector-mongodb/blob/4551cc0428a6954091981eb8e4861e7c1c9b37d8/test/decimal.test.js#L468-L471), hope it helps.

jannyHou (approved):
The proposal and PoC looks great!

SPIKE.md
bajtos: I feel it's important to get the developer experience right. If a user accidentally mis-configures their models, the framework should be helpful and tell them what they did wrong and how to fix it.

I am ok to work on this in a follow-up story, please make sure to include it the spike proposal. Should we start with a spike to better understand the complexities involved?

hacksparrow: >  If a user accidentally mis-configures their models, the framework should be helpful and tell them what they did wrong and how to fix it.

100% agreed, and best scoped in a separate story.

> Should we start with a spike to better understand the complexities involved?

I think so, validations are not going to be simple case of `if`-`else` when we have inferred and nested properties.

test/new-objectid.test.js
hacksparrow: Thanks Janny, added.

bajtos (changes requested):
Looks like we have a decent understanding of _where_ we want to get, what ObjectID-related behavior we want to have :+1:

Now we need to build a better understanding of _how_ to get there. What is the first step (the first task/user story)? What is the next one? Which tasks must be implemented one after other because of dependencies, and which can be worked on in parallel, potentially by more than just one person?

Ideally, the proposed plan should have enough details that the creation of follow-up/implementation GitHub issues will become mostly copy & paste, and that anybody else on the team or even from our community can pick up any task and contribute the changes, without having to fully understand the entire spike document.

hacksparrow:
@bajtos I will update the spike report with the follow up tasks.

hacksparrow:
@bajtos [added](https://github.com/strongloop/loopback-connector-mongodb/blob/d500f0ab6ccf2954c1be257025f021102b701fc9/SPIKE.md#follow-up-tasks).

> ## Follow-up Tasks

> 1. [ObjectID string interface](https://github.com/strongloop/loopback-connector-mongodb/issues/598)
> 2. [SPIKE: Connector level model validation](https://github.com/strongloop/loopback-connector-mongodb/issues/599)
> 3. [Primary key property not validated in `.create()`](https://github.com/strongloop/loopback-datasource-juggler/issues/1868)
> 4. [Refactor DAO methods](https://github.com/strongloop/loopback-datasource-juggler/issues/1867)

> 3 and 4 are somewhat related but not a requirement, they were discovered and created out of the [first spike](https://github.com/strongloop/loopback-next/issues/3456#issuecomment-673421592).

bajtos:
@hacksparrow thank you for opening follow-up tasks, they look mostly good 👍🏻 I remember that we were discussing that in order to support the desired behavior, where PKs are always represented as strings at juggler level, we need to remove `getDefaultIdType` to prevent juggler from changing model definition and always setting PK type of `ObjectID`.

https://github.com/strongloop/loopback-connector-mongodb/blob/4551cc0428a6954091981eb8e4861e7c1c9b37d8/lib/mongodb.js#L351-L353

Is this still relevant? Which story is covering this part? Should we create a new one, since this change seems relatively isolated? I think this change will require us to update many unit tests, so it may be best to plan it independently, to keep the size of https://github.com/strongloop/loopback-connector-mongodb/issues/598 smaller.

hacksparrow:
@bajtos sure, let me create a separate story for removing `getDefaultIdType()`. Thanks for the suggestion!

hacksparrow:
Created https://github.com/strongloop/loopback-connector-mongodb/issues/600.

Issue #600, opened 2020-09-25
Remove MongoDB.prototype.getDefaultIdType


hacksparrow:
Follow up of https://github.com/strongloop/loopback-next/issues/3456.

As a precursor to https://github.com/strongloop/loopback-connector-mongodb/issues/598, remove `MongoDB.prototype.getDefaultIdType`.

Tips: Consider return string instead of removing it which causes confusion. Add comments explain why we return string in mongodb connector.
- [ ] Update the function.
- [ ] Update and fix tests as well.

Issue #598, opened 2020-09-21
ObjectID string interface


hacksparrow:
All model properties defined as `{type: String, mongodb: {dataType: 'objectID'}` should be stored as `ObjectID` in the database but provide a string interface in the connector APIs. Meaning, string version of the `ObjectID` will be used in queries and results, `ObjectID` will not be directly exposed to users.

Any property not defined as `{type: String, mongodb: {dataType: 'objectID'}` will no be stored as `ObjectID`, even if the string looks like an ObjectID.

## Acceptance criteria

- [ ] Add implementation for `String` to `ObjectID` conversion in queries
- [ ] Add implementation for `ObjectID` to `String` conversion in results
- [ ] Refactor existing implementation wherever applicable
- [ ] Remove redundant code
- [ ] Update docs
- [ ] Add tests
- [ ] Add a "Long Term Support" section to README (see [strongloop/loopback](https://github.com/strongloop/loopback#module-long-term-support-policy) for inspiration) and also explain which connector versions are compatible with which LoopBack versions.

bajtos:
Important: we already have a similar implementation in place for `Decimal128` type. Going forward, we should have a single unified solution for dealing with string-based data-types (`ObjectID`, `Decimal128`), to ensure consistent handling of edge cases like nested properties in anonymous models, nested array items, etc.

hacksparrow:
Yes, in the proposed approach `ObjectID` and `Decimal128` coercion happen in the same block. It can be further optimized.

mikeevstropov:
@hacksparrow When you said
`Any property not defined as {type: String, mongodb: {dataType: 'objectID'} will no be stored as ObjectID, even if the string looks like an ObjectID.`
Does CLI will set `dataType` for relation automatically? It looks like i need to handle it manually each time when i generate the relation. :hushed:

paviln:
Hello,

When I use find on a repo, which model has a belongsTo:

  // Foreign key
  @belongsTo(
    () => User,
    {}, //relation metadata goes in here
    {
      // property definition goes in here
      type: 'string',
      required: true,
      mongodb: {dataType: 'ObjectId'},
    },
  )
  userId: string;

The result userId type is a object and not a string. This is annoying, since I want to use where neq..

The userId is stroed as ObjectId, if i change it to a string, I get a string value. But it should provide it as a string, even if it is a objectId?

PR #601, opened 2020-10-30
feat: string interface for ObjectID

hacksparrow:
Addresses https://github.com/strongloop/loopback-connector-mongodb/issues/600 and https://github.com/strongloop/loopback-connector-mongodb/issues/598.

## Checklist

- [ ] DCO (Developer Certificate of Origin) [signed in all commits](https://loopback.io/doc/en/contrib/code-contrib.html)
- [ ] `npm test` passes on your machine


test/id.test.js
```diff
@@ -247,164 +153,95 @@ describe('mongodb default id name', function() {
...
+});
 
-    it('should find model with ObjectID-like string id', async function() {
-      const user = await User.create({id: objectIdLikeString, name: 'John'});
-      user.id.should.not.be.an.instanceOf(ds.ObjectID);
-      user.id.should.eql(objectIdLikeString);
-      const found = await User.findById(objectIdLikeString);
-      found.toObject().name.should.equal('John');
-    });
+describe('non-generated ObjectId id', function() {
```
agnes512, line 177: Can we also have tests for non-autogenerated non-object-like string/number type id?

lib/mongodb.js
```diff
@@ -573,18 +555,34 @@ MongoDB.prototype.create = function(modelName, data, options, callback) {
...
   const idName = self.idName(modelName);
+  const propDef = self.getPropertyDefinition(modelName, idName);
+
+  if (hasDataType('objectid', propDef) && !propDef.generated && idValue) {
+    if (!ObjectIdValueRegex.test(idValue)) {
+      const err = new Error('Invalid ObjectId string.');
+      return callback(err);
+    }
+  } else if (propDef.generated && idValue) {
+    const err = new Error('Cannot specify value for autogenerated id.');
```
agnes512, line 569: I am a bit confused. Which case are we trying to catch here? Are we trying to reject if uses want to have auto-generated number/non-object-like string type id?

agnes512 (commented):
Thanks for taking care of the issue. I like how you rename those variables. They make the intention more clear. I left a few comments and questions.

lib/mongodb.js
hacksparrow: I am removing id checks for now, it is not compatible with `replaceOrCreate`.

lib/mongodb.js
```diff
@@ -2161,47 +2113,76 @@ function visitAllProperties(data, modelCtor, visitor) {
...
+      return propValue;
     } else {
-      propValue = coerceToObjectId(modelCtor, propDef, propValue);
+      return ObjectID(propValue);
+    }
+  } else if (propDef.id) {
+    if (typeof propValue !== 'number') {
+      // Autogenerated id
+      if (propDef.generated) {
+        return ObjectID(propValue);
```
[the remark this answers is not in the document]
hacksparrow: Since the id is generated by MongoDB, it will be ObjectId-like string in that block. `ObjectID()` will return the input as it is, if it cannot be converted to `ObjectId`.

test/id.test.js
hacksparrow: There are covered in other tests like https://github.com/strongloop/loopback-connector-mongodb/blob/64351c43e36c62699b3ecc79fe4844af4e66b1d7/test/id.test.js#L17 and https://github.com/strongloop/loopback-connector-mongodb/blob/major/string-objectid/test/mongodb.test.js#L255.

PR #602, opened 2020-11-03
Drop Juggler3 support

hacksparrow:
As part of https://github.com/strongloop/loopback-connector-mongodb/issues/598, we want to drop support for Juggler 3. I am opening a separate PR for cleaner and easier review.

## Checklist


test/objectid.test.js
```diff
@@ -139,8 +139,8 @@ describe('ObjectID', function() {
       const found = await Article.find({where: {title: 'arrayOfObjectID'}});
       // the type of the returned array is actually string even it's stored as ObjectIds in the db as expected
       found[0].xidArr.should.containDeep([
-        new ds.ObjectID(objectIDLikeString),
-        new ds.ObjectID(objectIDLikeString2),
+        objectIDLikeString,
```
hacksparrow, line 142: This is the correct  expected object, refer to the code comment above.

bajtos:
> The changes in this PR will be incorporated into the implementation PR https://github.com/strongloop/loopback-connector-mongodb/pull/601.

What is the reasoning for that? I prefer to land this pull request standalone. What problems do you envision if we drop support for juggler v3 in a standalone pull request?

README.md
```diff
@@ -35,27 +35,7 @@ const config = {
 
 <details><summary markdown="span"><strong>For LoopBack 3 users</strong></summary>
```
bajtos, line 36: Now that the content is just a single sentence, I feel we no longer need to hide it via `<details>` tag. Let's remove this line and perhaps make the next sentence in bold?

README.md
```diff
@@ -35,27 +35,7 @@ const config = {
...
-  "password": "mypassword",
-  "name": "mydb",
-  "user": "me",
-  "authSource" : "admin",
-  "connector": "mongodb"
-}
-```
-
-Edit `datasources.json` to add any other additional properties that you require.
+`loopback-connector-mongodb@6` is not compatible with LoopBack 3, `loopback-connector-mongodb@5.4.0` is the last supported version.
```
bajtos, line 38: When we release v7 in the future, we will have to remember to update this sentence. Let's find a way how to convey the information so that we don't have to update this line in the future.

Also note that we may publish `5.5.0` sometimes in the future if we need to backport a (security?) fix that's not a trivial patch.

Example content for inspiration:

> Starting from the version `6.0.0`, this connector is no longer compatible with LoopBack 3. Please use the latest `5.x` version in your LoopBack 3 applications.

bajtos (changes requested):
I have few comments to address, see below.

Please add a LTS section to README with a table showing major versions and their support status. See https://github.com/strongloop/strong-error-handler#supported-versions for an example. IMO, we should add a new column to the table to list which LoopBack versions are supported by each connector major version.

test/objectid.test.js
bajtos: I feel this change is unrelated to dropping support for juggler v3, can we please remove it from this PR and land it as part of your ObjectID refactor?

README.md
dhmlau: We get questions periodically about which version of LB, juggler and connector that work together. Perhaps we can mention the juggler version as well (better in a table form)?

hacksparrow:
> What is the reasoning for that? I prefer to land this pull request standalone. What problems do you envision if we drop support for juggler v3 in a standalone pull request?

The reason I can think of is something like what you said:

> Also note that we may publish 5.5.0 sometimes in the future if we need to backport a (security?) fix that's not a trivial patch.

If we are sure about not supporting Juggler 3 at all after 5.4, we can land this PR on master as a standalone PR.

test/objectid.test.js
hacksparrow: There is some sort of connection between the Juggler 3 tests and this particular test case, removing the Juggler 3 tests make this test case fail (as expected).

README.md
```diff
@@ -33,32 +44,6 @@ const config = {
 };
 ```
 
-<details><summary markdown="span"><strong>For LoopBack 3 users</strong></summary>
```
agnes512, line 36: Since `6.x` will just for LB4, do we want to remove all the collapsible sections in REAME?

README.md
bajtos: > We get questions periodically about which version of LB, juggler and connector that work together. Perhaps we can mention the juggler version as well (better in a table form)?

That would be awesome 👏🏻 

I am ok to leave it out of scope of this pull request though, if it helps to get the changes landed faster.

README.md
bajtos: > Since `6.x` will just for LB4, do we want to remove all the collapsible sections in README?

Yes please :+1:

bajtos:
> If we are sure about not supporting Juggler 3 at all after 5.4, we can land this PR on master as a standalone PR. Picture the support matrix in this case, are we fine with it?

Here is the process to follow:

1. Just before we land this PR (the first breaking change since 5.0), we create a new long-living branch `5.x` from the last commit before the breaking change.
2. As part of this PR, we change  package version to something like `6.0.0-1`, this will be committed to `master`. It will also allow us to start publishing 6.0 pre-releases (although we should use a different npm tag if we decide to do that, see the step 3 for related info).
2. Until we publish 6.0, we should back-port all backwards compatible changes (features, bug fixes) from `master` to `5.x`. New releases will be published from `5.x`, not from `master`.
3. When we are ready for `6.0`:
  - Update `publishConfig.tag` to `lts` in the branch `5.x`, see e.g. [`loopback@2.x/package.json`](https://github.com/strongloop/loopback/blob/80cb175105ee542334cda0c0ce881cf9d454d2b0/package.json#L4-L6)
  - Publish `6.0.0` from `master`
4. Once 6.0 is published, only backwards-compatible bug fixes are back-ported to `5.x` branch.

This is a well-tried approach we had already followed in the past, e.g. while working on LB 3.0 release. No need to reinvent the wheel here.

Regarding the support matrix: 

- Until `6.0` is released,
  - The version `5.x` is "Current", but lives in `5.x` branch (not `master`). It supports both juggler v3 and v4.
  - The `master` branch is tracking the upcoming `6.0` release.  It supports juggler v4 only.
- After we release `6.0`:
  - The version `6.0` becomes "Current", it supports juggler v4 only.
  - The version `5.x` moves to "Active LTS", it supports juggler v3 and v4.
  - The version `4.x` moves to "Maintenance LTS", it supports juggler v3 and v4. (Maybe v2 too? I am not sure.)
  - All older versions reach end-of-life

README.md
```diff
@@ -14,6 +14,17 @@ This installs the module from npm and adds it as a dependency to the application
...
+## Supported versions
+
+**Starting from the version 6.0.0, this connector is no longer compatible with LoopBack 3. Please use the latest 5.x version in your LoopBack 3 applications.**
+
+This module adopts the [Module Long Term Support (LTS)](http://github.com/CloudNativeJS/ModuleLTS) policy, with the following End Of Life (EOL) dates:
+
+| Version    | Status          | Published | EOL                  | LoopBack | Juggler  |
+| ---------- | --------------- | --------- | -------------------- | ---------|----------|
+| 6.x        | Current         | Nov 2020  | Apr 2023 _(minimum)_ | 4        | 4.x      |
+| 5.x        | Active LTS      | Jun 2019  | Aug 2021             | 3, 4     | 3.x, 4.x |
```
bajtos, line 26: Much better! 👏🏻 

Let's add a row for the 4.x version of the connector please, it should be listed as "Maintenance LTS".

test/objectid.test.js
bajtos: I see. If the behavior is fixed now, then I think the comment no longer makes sense and should be removed?

bajtos (commented):
@hacksparrow this PR looks almost good to be landed, can you please address the recent comments posted by Agnes and Diana?

README.md
bajtos: AFAICT, the support matrix table introduced by d0bb62d is already listing supported LoopBack versions, so I think this comment has been resolved.

bajtos:
@slnode test please

README.md
hacksparrow: What is the EOL of 4.x?

test/objectid.test.js
hacksparrow: > I think the comment no longer makes sense and should be removed?

Which comment should be removed?

In case it was not clear, Juggler 3 tests were interfering in some way and causing the test case to pass when it should have actually failed because the array will contain string values not ObjectIds.

README.md
bajtos: We are following https://github.com/CloudNativeJS/ModuleLTS.

> When a version of Node.js enters LTS, whichever is the latest major version (X.y.z) of an LTS adopting module at that time will be supported for at least security and critical fixes for the lifetime of that version of Node.js.

loopback-connector-mongodb 5.x was released in Jun 2019, at which time the latest LTS version of Node.js was 10.x.  Before our release, people were using loopback-connector-mongodb 4.x with Node.js 10.x, thus we need to support connector 4.x for at least as long as Node.js will support 10.x - that means 2021-04-30 - see https://github.com/nodejs/Release

As for loopback-connector-mongodb 5.x, we need to support it until the current Node.js LTS version 14.x goes end of life, which means April 2023. Please update the row in the table accordingly.

README.md
```diff
@@ -14,6 +14,18 @@ This installs the module from npm and adds it as a dependency to the application
...
+
+**Starting from the version 6.0.0, this connector is no longer compatible with LoopBack 3. Please use the latest 5.x version in your LoopBack 3 applications.**
+
+This module adopts the [Module Long Term Support (LTS)](http://github.com/CloudNativeJS/ModuleLTS) policy, with the following End Of Life (EOL) dates:
+
+| Version    | Status               | Published | EOL                  | LoopBack | Juggler  |
+| ---------- | -------------------- | --------- | -------------------- | ---------|----------|
+| 6.x        | Current              | Nov 2020  | Apr 2023 _(minimum)_ | 4        | 4.x      |
+| 5.x        | Active LTS           | Jun 2019  | Aug 2021             | 3, 4     | 3.x, 4.x |
+| 4.x        | Maintenance LTS      | Mov 2018  | Aug 2021             | 3        | 3.x      |
```
bajtos, line 27: Please update EOL dates per the discussion in https://github.com/strongloop/loopback-connector-mongodb/pull/602#discussion_r531096785

Also I believe the connector version 4.x supports LB4 and juggler v4 too, it was published after LB4 went GA on Oct 10, 2018.

```suggestion
| 6.x        | Current              | Nov 2020  | Apr 2023 _(minimum)_ | 4        | 4.x      |
| 5.x        | Active LTS           | Jun 2019  | Apr 2023             | 3, 4     | 3.x, 4.x |
| 4.x        | Maintenance LTS      | Nov 2018  | Apr 2021             | 3,4      | 3.x,4.x  |
```

bajtos (commented):
Almost there!

bajtos:
Since we are not using semantic commits in this repository, but still building release notes from commit messages, I am proposing to use the following commit message when you get to git history cleanup before landing:

```
[SEMVER-MAJOR] Drop support for LoopBack 3.x
```

The goal is to have a message that uses the language that a typical LoopBack user will understand and clearly communicate the fact that we are no longer supporting LB 3 and it's a breaking change.

agnes512 (approved):
👏

README.md
```diff
@@ -14,6 +14,18 @@ This installs the module from npm and adds it as a dependency to the application
...
+
+**Starting from the version 6.0.0, this connector is no longer compatible with LoopBack 3. Please use the latest 5.x version in your LoopBack 3 applications.**
+
+This module adopts the [Module Long Term Support (LTS)](http://github.com/CloudNativeJS/ModuleLTS) policy, with the following End Of Life (EOL) dates:
+
+| Version    | Status               | Published | EOL                  | LoopBack | Juggler  |
+| ---------- | -------------------- | --------- | -------------------- | ---------|----------|
+| 6.x        | Current              | Nov 2020  | Apr 2023 _(minimum)_ | 4        | 4.x      |
+| 5.x        | Active LTS           | Jun 2019  | Apr 2023             | 3, 4     | 3.x, 4.x |
+| 4.x        | Maintenance LTS      | Nov 2018  | Apr 2021             | 3, 4      | 3.x, 4.x  |
```
bajtos, line 27: Nipick - remove one space after `3, 4` in the last row to keep the the very last table cell indented with the rest.

```suggestion
| 6.x        | Current              | Nov 2020  | Apr 2023 _(minimum)_ | 4        | 4.x      |
| 5.x        | Active LTS           | Jun 2019  | Apr 2023             | 3, 4     | 3.x, 4.x |
| 4.x        | Maintenance LTS      | Nov 2018  | Apr 2021             | 3, 4     | 3.x, 4.x  |
```

test/objectid.test.js
bajtos: I was referring to the following comment on line 140:

```js
// the type of the returned array is actually string even it's stored as ObjectIds in the db as expected
```

As I am re-reading it now, the comment seems valid, so I am fine to keep it.

bajtos (approved):
Code changes LGTM with one nitpick to consider.

Please remember to squash all commits into a single one with a descriptive commit message, see https://github.com/strongloop/loopback-connector-mongodb/pull/602#issuecomment-734691808

> Since we are not using semantic commits in this repository, but still building release notes from commit messages, I am proposing to use the following commit message when you get to git history cleanup before landing:
> 
> ```
> [SEMVER-MAJOR] Drop support for LoopBack 3.x
> ```
> 
> The goal is to have a message that uses the language that a typical LoopBack user will understand and clearly communicate the fact that we are no longer supporting LB 3 and it's a breaking change.

hacksparrow:
@bajtos extra spaces removed, commits squashed, commit message changed to `[SEMVER-MAJOR] Drop support for LoopBack 3.x`.

hacksparrow:
@slnode test please

PR #613, opened 2021-01-21
ci: add GH action and remove Travis CI integration

agnes512:
Notice that we were using mongodb 3.2 for Travis. That version runs into connection errors in GitHub action. This PR uses the latest stable version of mongodb (4.4) for Github action.

## Checklist


.github/workflows/continous-integration.yaml
```diff
@@ -0,0 +1,37 @@
+name: Continous Integration
```
dhmlau, line 1: typo

```suggestion
name: Continuous Integration
```

.github/workflows/continous-integration.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Start MongoDB
+      uses: supercharge/mongodb-github-action@1.3.0
+      with:
+        mongodb-version: ${{ matrix.mongodb-version }}
+
+    - run: npm install
+    - run: npm test
+      env:
+        CI: true
+
```
dhmlau, line 37: It's complaining no new line. Is it because there's extra space on the last line?

dhmlau (commented):
I have limited knowledge on GitHub Actions, but it LGTM. Please see my minor comments.

.github/workflows/continous-integration.yaml
agnes512: Yes you're right. Fixed!

.travis.yml
```diff
@@ -1,21 +0,0 @@
...
-      - ubuntu-toolchain-r-test
-      - mongodb-3.2-precise
-    packages:
-      - g++-4.8
-      - mongodb-org-server
-      - mongodb-org-shell
-services:
-  - mongodb
-
-after_success: npm run coverage
```
bajtos, line 21: I believe this command is publishing code coverage data to https://coveralls.io/github/strongloop/loopback-connector-mongodb. Can you please preserve this step in the new GHA workflow too?

I think it should be enough to add a new step at the end of the file:

```
    - run: npm install
    - run: npm test
      env:
        CI: true
    - run: npm run coverage
```

bajtos (approved):
Great job, @agnes512! I have one comment to address, PTAL 👇🏻 

Considering that our CI is broken right now, I think it's ok to land this change as-is even if it's breaking code coverage, and then restore Coveralls integration in a follow-up pull request.

agnes512:
I was investigating adding coveralls to the workflow. But GitHub action seems disappeared 🤔 I am only seeing jenkins running. Not sure if I messed up something.

agnes512:
@bajtos I tried to enable coveralls like https://coveralls.io/builds/32304908. But I don't know how to do it like Travis does. Seems like Travis has some repo token that coveralls.io would use. Not sure how this can be done with `nyc` + `github action`. I tried all possible solutions that I could fine but still still have no luck. 

Also, seems like the `nyc report | coveralls` still provides some coverage report:
image

and it has `.nyc_output` locally, which contains such reports:
image


Not sure if it's possible to have pages like coveralls.io for our case. Would you want me to invest more? 

I also noticed that when we changed from Travis to Github action, we dropped the coveralls report for loopback-next. https://github.com/strongloop/loopback-next/pull/5110 Again, not sure how complicated it would be. But we might want to have the report for loopback-next. Please let me know if you have any concerns. Thanks.

PR #626, opened 2021-05-08
feat: add transactions support

darknos:
add base transaction support for for mongodb v4+ shard and replica sets

Tests were copied from https://github.com/strongloop/loopback-connector/blob/master/test/transaction.test.js but skip by default, because we need replica set for tests.

I use run-rs for testing
```
run-rs -v 4.2.0 --host localhost --portStart 27000
```

just remove `.skip` from test/transaction.test.js to run it

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "loopback-connector-mongodb",
   "version": "5.5.0",
+  "publishConfig": {
+    "tag": "lts"
+  },
   "description": "The official MongoDB connector for the LoopBack framework.",
   "engines": {
     "node": ">=10"


dhmlau:
@darknos, thanks for the PR. Could you please add tests as well? Thanks.

darknos:
@dhmlau added. (I missed the file when comiited last time)

darknos:
> @darknos, thanks for the PR. Could you please add tests as well? Thanks.

any news?

taicho:
@dhmlau Also looking for news on this...

package.json
```diff
@@ -48,6 +48,8 @@
   "devDependencies": {
     "benchmark": "^2.1.4",
     "bluebird": "^3.5.4",
+    "chai": "^4.3.4",
```
marioestradarosa, line 51: We usually use mocha , not sure why you added another package here, please explain ?

marioestradarosa (changes requested):
Not so sure if there was a **chai** feature you required that mocha doesn't have it.

lib/mongodb.js
```diff
@@ -1488,15 +1488,26 @@ MongoDB.prototype.count = function count(modelName, where, options, callback) {
     debug('count', modelName, where);
   }
   where = self.buildWhere(modelName, where, options) || {};
-  const method = Object.keys(where).length === 0 ? 'estimatedDocumentCount' : 'countDocuments';
```
marioestradarosa, line 1491: Why not use 
```ts
options = buildOptions({}, options);
const method = (Object.keys(where).length === 0  && !options.session) ? 'estimatedDocumentCount' : 'countDocuments';
```
To simplify/remove your new condition?

lib/mongodb.js
```diff
@@ -2015,6 +2026,64 @@ MongoDB.prototype.ping = function(cb) {
...
 };
 
+MongoDB.prototype.beginTransaction = function(isolationLevel, cb) {
+  // TODO: think about how to convert READ_COMMITED, etc. to transactionOptions
+  const transactionOptions = {
+    readPreference: 'primary',
+    readConcern: {level: 'local'},
+    writeConcern: {w: 'majority'},
+  };
+  if (typeof isolationLevel === 'object') {
```
marioestradarosa, line 2036: for nullSafety , you may use. Keep in mind that typeof returns true when checked against null
```ts
 if (isolationLevel instanceof Object) {
```

lib/mongodb.js
```diff
@@ -2015,6 +2026,64 @@ MongoDB.prototype.ping = function(cb) {
...
+  tx.abortTransaction(function(err) {
+    if (err) {
+      // end session in any case
+      tx.endSession(null, function(error) {
+        if (err) return cb(err);
+        if (error) return cb(error);
+        cb();
+      });
+    } else {
+      tx.endSession(null, function(error) {
```
marioestradarosa, line 2072: You duplicate this code. Not so sure if it will work, but I suggest you remove the first ```if (err) {``` because you are ending the session anyway. Also remove the inner ```else {``` to simplify your code.

package.json
darknos: I copied transactions tests from loopback-cinnector without any changes but it uses chai. I hope it is acceptable.

darknos: https://github.com/strongloop/loopback-connector/blob/master/test/transaction.test.js

lib/mongodb.js
darknos: execute with options has different signature for estimatedDocumentCount and countDocuments. check the code below.

lib/mongodb.js
darknos: old school. :) I'll change it ot instanceOf

lib/mongodb.js
darknos:
```
MongoDB.prototype.rollback = function(tx, cb) {
  tx.abortTransaction(function(err) {
    tx.endSession(null, function(error) {
      if (err) return cb(err);
      if (error) return cb(error);
      cb();
    });
  });
};
```

like this?

package.json
marioestradarosa: Can you help us standardize it here?. I double checked and connectors (ie: mysql, postgresql and this mongodb) already have mocha in the package.json so I guess it should work with it.   Thanks for your patience , this is a great addition to the mongo DB connector.

lib/mongodb.js
marioestradarosa: Correct!, same for ```MongoDB.prototype.commit```

lib/mongodb.js
darknos: done

lib/mongodb.js
darknos: done

package.json
darknos: I replaced expect with should in tests

package.json
```diff
@@ -48,6 +48,7 @@
   "devDependencies": {
     "benchmark": "^2.1.4",
     "bluebird": "^3.5.4",
+    "chai-as-promised": "^7.1.1",
```
marioestradarosa, line 51: You may remove this unused package ```chai-as-promised``` , I pulled your changes and all local tests are passing without it.

test/transaction.test.js
```diff
@@ -0,0 +1,235 @@
+// Copyright IBM Corp. 2013,2019. All Rights Reserved.
+// Node module: loopback-connector-mongodb
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+const should = require('./init.js');
+const chaiAsPromised = require('chai-as-promised');
```
marioestradarosa, line 8: You may remove this line, chai-as-promised is not needed. I copied your PR and all tests passed.

marioestradarosa (changes requested):
@darknos great Job, many thanks for this PR. I only had two minor comments about the chai-as-promise package. I fetched your PR locally in my machine and all works perfectly without these two references for this package. 

Also, please squash your commits to keep the history tidy and easier to digest.

darknos:
@marioestradarosa Hi, could you help to understand what the problem with PR Builder?

I did all requested changes but don't have access to Jenkins to check the problem

achrinza:
Hi @darknos, apologies for not noticing your query earlier.

As part of LoopBack 4 joining the OpenJS Foundation, we're in the midst of migrating the CI infrastructure from Jenkins to GitHub Actions and Travis CI.

Let me see if I can sort it out from my end and get back to you.

achrinza:
Yikes! That wasn't supposed to happen... Really sorry about that :bow: 

I've re-pushed the commit under #633.

PR #637, opened 2021-08-16
fix: handle url default db name

### LINKED ISSUE #610: db name is always admin if url is used without db
## Steps to reproduce

- Set `url` parameter in connection properties to `mongodb://localhost:27017`
- Set `database` parameter to `mydb`

## Current Behavior

It will use `admin` as database

## Expected Behavior

It should use `mydb` as database

## Additional information

In the connector it checks if the `dbName` is in the url ([mongodb.js#L338](https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L338)) and if not it should use `settings.database` but this is does not quite work because in the url will always be a database, the mongo driver will always set `admin` as default in the url if no database is specified ([url_parser.js#L377](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/url_parser.js#L377))

achrinza:
closes #610

Signed-off-by: Rifa Achrinza <25147899+achrinza@users.noreply.github.com>

## Checklist

- New tests added or existing tests modified to cover all changes

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -325,6 +325,15 @@ MongoDB.prototype.connect = function(callback) {
       if (callback) callback(err);
     }
 
+    const urlObj = new URL(self.settings.url);
+
+    if ((urlObj.pathname === '' ||
+        urlObj.pathname.split('/')[1] === '') &&
+        typeof self.settings.database === 'string') {
+      urlObj.pathname = self.settings.database;
+      self.settings.url = urlObj.toString();
+    }
+
     new mongodb.MongoClient(self.settings.url, validOptions).connect(function(
       err,
       client,


lib/mongodb.js
```diff
@@ -325,6 +325,15 @@ MongoDB.prototype.connect = function(callback) {
       if (callback) callback(err);
     }
 
+    const urlObj = new URL(self.settings.url);
+
+    if (urlObj.pathname !== '' &&
+        urlObj.pathname.split('/')[1] === '' &&
+        typeof self.settings.database === 'string') {
```
nflaig, line 332: @achrinza if `mongodb://localhost:27017` is used in `settings.url` the `pathname` will be `''` and in this case we should check if `settings.database` is set and set the database name as path name

```suggestion
    if ((urlObj.pathname === '' ||
        urlObj.pathname.split('/')[1] === '') &&
        typeof self.settings.database === 'string') {
```

nflaig (changes requested):

lib/mongodb.js
achrinza: Ah, Iooks like I initially misunderstood when the default db name is injected by mongo driver. Thanks :+1:

nflaig (approved):

agnes512 (approved):

PR #794, opened 2025-07-01
fix: include options parameter in createIndex method

digorgonzola:
**Problem:** When defining custom index names in model settings, the connector was ignoring the specified `name` option and falling back to MongoDB's default naming convention (e.g., "archived_1" instead of "archived").

**Root Cause:** The `createIndex()` call was missing the `options` parameter, so custom index options including the `name` field were not being passed to MongoDB.

**Solution:** Updated the `createIndex()` call to include the `index.options` parameter:

## Checklist

- [X] DCO (Developer Certificate of Origin) [signed in all commits](https://loopback.io/doc/en/contrib/code-contrib.html)
- [x] `npm test` passes on your machine (fails 1 test but this is unrelated)
- [X] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [X] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -2228,7 +2228,7 @@ MongoDB.prototype.autoupdate = function(models, cb) {
               debug('createIndex: ', index);
             }
             const callbackCreateIndex = util.callbackify(() => self
-              .collection(modelName).createIndex(index.fields || index.keys));
+              .collection(modelName).createIndex(index.fields || index.keys, index.options));
             callbackCreateIndex(indexCallback);
           },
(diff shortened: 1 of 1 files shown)


rahulrkr08:
Adding +1 for this.  xxx-alpha.5 version is breaking migration (index creation).

dhmlau:
@digorgonzola, thanks for your contribution. Your PR has landed! 🎉

PR #901, opened 2026-07-01
fix: #900 replace callback implementation by the promise one

### LINKED ISSUE #900: commit / rollback are stuck forever
Commit and rollback methods get stuck forever.
They rely on the callback form of the underlying driver that was removed in recent MongoDB drivers in favor of the promise one. So with mongodb driver version 5 and above (currently referenced is ^5.9.2), the callback never gets called, and the call never ends.

This bug affects all versions where mongodb driver has been upgraded. This includes the 7.X ones.

## Steps to reproduce

Simply un-skip the transaction test and run it.

## Current Behavior

Tx commit() is stuck

## Expected Behavior

It should answer ;-)

## Link to reproduction sandbox

Just run the transaction test 

## Additional information

Fixing the implementation can be done by relying on promises, like:
```javascript
MongoDB.prototype.commit = function (tx, cb) {
  tx.commitTransaction()
    .then(() => tx.endSession())
    .then(() => cb())
    .catch((err) => cb(err));
};

MongoDB.prototype.rollback = function (tx, cb) {
  tx.abortTransaction()
    .then(() => tx.endSession())
    .then(() => cb())
    .catch((err) => cb(err));
};
```

## Related Issues

N/A

vvandens:
Starting from version 5, the mongodb node driver removed callback support in favor of promises only. This PR fixes the commit and rollback implementation of the connector.
No extra test needed. It is already covered by the transaction one (that is skipped...).

Fixes #900 

## Checklist

### DIFF
--- a/lib/mongodb.js
+++ b/lib/mongodb.js
@@ -2349,23 +2349,17 @@ MongoDB.prototype.beginTransaction = function(isolationLevel, cb) {
 };
 
 MongoDB.prototype.commit = function(tx, cb) {
-  tx.commitTransaction(function(err) {
-    tx.endSession(null, function(error) {
-      if (err) return cb(err);
-      if (error) return cb(error);
-      cb();
-    });
-  });
+  tx.commitTransaction()
+    .then(() => tx.endSession())
+    .then(() => cb())
+    .catch((err) => cb(err));
 };
 
 MongoDB.prototype.rollback = function(tx, cb) {
-  tx.abortTransaction(function(err) {
-    tx.endSession(null, function(error) {
-      if (err) return cb(err);
-      if (error) return cb(error);
-      cb();
-    });
-  });
+  tx.abortTransaction()
+    .then(() => tx.endSession())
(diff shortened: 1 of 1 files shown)


achrinza (approved):
Thanks for the quick PR turnaround! All LGTM 👍 

I'll give the other maintainers an opportunity to review this PR if they wish. Otherwise, we'll proceed to merge this after ~one day.

vvandens:
Perfect.
It would need a backport to the 7.x branch of course.

achrinza:
We can consider backporting to v7 if there's a use-case for it.

From a cursory check, v8's only breaking change is dropping support for Node.js v20, which has already reached End of Life.

Is there any particular blocker that's preventing an upgrade to v8?

vvandens:
Not our side. We can definitely upgrade to V8.
But it's a pretty big issue IMHO. It makes the whole 7.x versions unusable as soon as you use transactions. As of now, we have monkey-patched the prototype at runtime.

